PR #38, opened 2013-05-04
Google Code Issue 157: Add "escape invisible characters" option

gsnedders:
> Reported by @fantasai, Jul 27, 2010 
> Having invisible characters in the source code can be confusing to someone who's trying to figure out what's going on. Adding an escape_invisible option would make those characters visible in the source code.
> 
> The attached patch implements an escape_invisible option. The list of invisible characters is probably incomplete for this iteration, but you get the idea. It depends on the patch in  issue 156 .

### DIFF
--- a/html5lib/serializer/htmlserializer.py
+++ b/html5lib/serializer/htmlserializer.py
@@ -167,7 +167,7 @@ class HTMLSerializer(object):
     def encode(self, string):
         assert(isinstance(string, text_type))
         if self.escape_invisible:
-            text = utils.escapeInvisible(text, self.escape_invisible == 'named')
+            string = utils.escapeInvisible(string, self.escape_invisible == 'named')
         if self.encoding:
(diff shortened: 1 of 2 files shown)


gsnedders:
My preference would be something based on `unicodedata` and blacklisting General Category C* (though that has the problem that you'll end up blacklisting different sets of characters depending on the Python version and the Unicode version, and generating that set is expensive and hence likely should be precomputed at dist build-time, and likely needs to be represented as a segment tree rather than a set of millions of characters out of concern for memory consumption).

We also need to be careful on narrow Python builds and make sure we don't encode surrogate pairs, as `\uD800\uDC00` needs to end up unchanged.

It's also notable that AFAICT the origin reason for this patch no longer holds true (the CSS testsuite build system is basically a historical artefact now and hasn't used an html5lib fork with this for years).

PR #80, opened 2013-06-25
Using frozenset with lists to avoid problems.

tonylopes:
It makes more sense to pass lists to frozensets since tuples need special syntax to be interpreted correctly in this context. For instance, the line  "hr": frozenset(("noshade")) has a bug. It won't produce a frozenset with "noshade" but with ['a', 'e', 'd', 'h', 'o', 'n', 's'] because if you forget to add a comma at the end, the tuple won't be an iterable but the string is it interprets is.

### DIFF
--- a/html5lib/constants.py
+++ b/html5lib/constants.py
@@ -298,7 +298,7 @@ namespaces = {
     "xmlns": "http://www.w3.org/2000/xmlns/"
 }
 
-scopingElements = frozenset((
+scopingElements = frozenset([
     (namespaces["html"], "applet"),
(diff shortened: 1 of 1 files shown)


gsnedders:
Tests: `<hr n>`, `<hr n=b>`, `<hr n=n>`, `<hr noshade>`, `<hr noshade=b>`, `<hr noshade=noshade>`.

gsnedders:
Rebased and merged into master as  a98f8a65b4514207d6c8f90cad8d4ec9f8ba6b41

PR #93, opened 2013-07-19
Fix #85: remove localization of error messages.

### LINKED ISSUE #85: Localizable error messages
Currently almost all our error messages are passed through gettext. Do we actually want localizable error messages? To me, at least, this seems an anti-feature.

cc/ @ambv and @jgraham

gsnedders:
It doesn't make sense to localize error messages, given they are purely technical in content. (#85, given GitHub doesn't notice titles.)


gsnedders:
Merged as e1e8d89c4bfe5c22fd3918a82e3215800d253752

PR #106, opened 2013-08-16
removed duplicate dispatch of noframes in inBodyPhase startTagHandler

### LINKED ISSUE #100: inBodyPhase incorrectly duplicating tag names => method dispatch
In the parser, the startTagHandler is sending "noframes" to startTagProcessInHead, but also it's listed under startTagRawtext.

Similarly the endTagHandler in the inBody phase incorrectly lists "dl" as one of the endTagListItem triggers; it's already being consumed by endTagBlock.

Since in both cases it's the first method that's correctly picked up, so doesn't actually have any result, but should be removed anyway. 

Thanks.

lilbludevil:
now dispatches noframes to startTagRawtext. fixes #100

### DIFF
--- a/html5lib/html5parser.py
+++ b/html5lib/html5parser.py
@@ -868,7 +868,7 @@ def getPhases(debug):
             self.startTagHandler = utils.MethodDispatcher([
                 ("html", self.startTagHtml),
                 (("base", "basefont", "bgsound", "command", "link", "meta",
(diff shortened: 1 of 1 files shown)


gsnedders:
To stop me in the future from getting confused: this has no blackbox observable affect.

gsnedders:
Merged as f1176e18966089e3b48659452aa951f2de285ab3

PR #110, opened 2013-08-27
Move the sanitizer to purely be a filter.

### LINKED ISSUE #72: Sanitizing filter broken in 0.90
http://code.google.com/p/html5lib/issues/detail?id=162

Reported by gdr@garethrees.org, Oct 10, 2010 
> DESCRIPTION
> 
> Consider the following interaction with html5lib 0.90:
> 
> ```
>     >>> from html5lib import html5parser, serializer, treebuilders, treewalkers
>     >>> p = html5parser.HTMLParser(tree = treebuilders.getTreeBuilder('dom'))
>     >>> dom = p.parse("""<body onload="sucker()">""") 
>     >>> s = serializer.htmlserializer.HTMLSerializer(sanitize = True)
>     >>> ''.join(s.serialize(treewalkers.getTreeWalker('dom')(dom)))
>     u'<body onload=sucker()>'
> ```
> 
> This is clearly incorrect: the onload attribute should have been removed by the sanitizer during the serialization.
> 
> ANALYSIS
> 
> The problem is that there are two sanitizers: a tokenizing sanitizer in html5lib.sanitizer, and a sanitizing filter in html5lib.filter.sanitizer. To avoid duplication of code, these two sanitizers inherit from the class HTMLSanitizerMixin and both call that class's function sanitize_token.
> 
> Unfortunately, the format of tokens differs between tokenization and filtering. During tokenization, a token looks like this:
> 
> ```
>     >>> from html5lib import tokenizer
>     >>> next(iter(tokenizer.HTMLTokenizer("""<body onload="sucker()">""")))
>     {'selfClosing': False, 'data': [[u'onload', u'sucker()']], 'type': 3, 'name': u'body', 'selfClosingAcknowledged': False}
> ```
> 
> But during filtering, tokens look like this:
> 
> ```
>     >>> list(iter(treewalkers.getTreeWalker('dom')(dom)))[3]
>     {'namespace': u'http:/​/​www.w3.org/​1999/​xhtml', 'type': 'StartTag', 'name': u'body', 'data': [(u'onload', u'sucker()')]}
> ```
> 
> When the sanitizing filter passes its token to the sanitize_token method of HTMLSanitizerMixin, nothing happens, because sanitize_token is expecting 'type' to be an integer.
> 
> OBSERVATION
> 
> Having two very similar but subtly different data formats for the same data type is dangerous: how many other incompatibilities are there?
> 
> WORKAROUND
> 
> I am working around this problem as follows: when I need to apply a sanitizing filter to a DOM tree, instead I do the following:
> 
> 1. Serialize the DOM to HTML without sanitization.
> 2. Re-parse the HTML from step 1, using the sanitizing tokenizer.

gsnedders:
This drops support for the tokenizing side of thing, which is sadly the only side that works in previous releases. Fixes #72. See all discussion there.


gsnedders:
I think when squashed this is now ready to land.

html5lib/filters/sanitizer.py
```diff
@@ -1,12 +1,872 @@
...
+
+            # Remove forbidden attributes
+            for to_remove in (attr_names - self.allowed_attributes):
+                del token["data"][to_remove]
+                attr_names.remove(to_remove)
+
+            # Remove attributes with disallowed URL values
+            for attr in (attr_names & self.attr_val_is_uri):
+                assert attr in attrs
+                # I don't have a clue where this regexp comes from or why it matches those
```
mgilson: It looks like this regex is trying to remove:

```
`                Literal backtic...
\x00-\x20        Ascii table characters starting at NULL up to and including Space -- Mostly non-printable characters...
\x7f-\xa0        Bytes with values 127 -> 160.  More non-printable characters...
\s               Whitespace
```

gsnedders: My assumption is it's something to do with legacy attribute parsing in old UAs, for whom such characters can alter parsing.

PR #126, opened 2013-12-09
Parser changes for the new ruby model

darobin:
As per http://darobin.github.io/html-ruby/. This change is being landed in the W3C specification, apply with caution.

### DIFF
--- a/html5lib/html5parser.py
+++ b/html5lib/html5parser.py
@@ -901,7 +901,7 @@ def getPhases(debug):
                 ("iframe", self.startTagIFrame),
                 (("noembed", "noframes", "noscript"), self.startTagRawtext),
                 ("select", self.startTagSelect),
-                (("rp", "rt"), self.startTagRpRt),
+                (("rp", "rt", "rb", "rtc"), self.startTagRpRtRbRtc),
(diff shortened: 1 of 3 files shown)


html5lib/html5parser.py
```diff
@@ -901,7 +901,7 @@ def __init__(self, parser, tree):
                 ("iframe", self.startTagIFrame),
                 (("noembed", "noframes", "noscript"), self.startTagRawtext),
                 ("select", self.startTagSelect),
-                (("rp", "rt"), self.startTagRpRt),
+                (("rp", "rt", "rb", "rtc"), self.startTagRpRtRbRtc),
```
annevk: This is wrong. You need two distinct groups. Or you need to also branch the `self.tree.openElements[-1].name != "ruby"` line further. What you're doing now doesn't match other implementations.

darobin: I'm not sure that this PR is up to date with the more recent changes to the spec anyway.

annevk: It isn't.

gsnedders:
I'm just going to close this as outdated and deal with this with the larger spec update.

PR #135, opened 2013-12-28
convert to setuptools and list all dependencies explicitly

mindw:


gsnedders:
The setuptools stuff no longer applies; the docs stuff has landed as 46dae3d9f471468da5890803e80115db246ea6b6

Issue #127, opened 2013-12-17
http.client.HTTPResponse is not playing nice (Python 3.x)


SimonSapin:
This gives the empty string on Python 3.x

``` python
from urllib.client import urlopen
html5lib.serialize(html5lib.parse(urlopen(
    'http://html5lib.readthedocs.org/en/latest/')))
```

The cause is a bug in Python: http://bugs.python.org/issue20007 , bug given CPython’s release cycle timeline I would like to have a work-around in html5lib.

The bug is triggered here:
https://github.com/html5lib/html5lib-python/blob/e269a2fd0aafcd83af7cf1e65bba65c0e5a2c18b/html5lib/inputstream.py#L122

Unfortunately, the only work-around I can think of (adding a special case `if isinstance(source, http.client.HTTPResponse): isUnicode = False` ) is very ugly.

PR #136, opened 2013-12-31
WIP More general fix for #127 with addinfourl

SimonSapin:
**Do not merge yet**, this is incomplete.

As discussed in #134 I changed this to avoid `.read(0)` entirely and pass a first chunk to `HTMLUnicodeInputStream` and `HTMLBinaryInputStream`, but then I get lost in their implementation and I don’t know what to do with that first chunk.

Surprisingly, most tests still passed because `.seek(0)` is used in some places. I added a failing test with a non-seekable input.

### DIFF
--- a/html5lib/inputstream.py
+++ b/html5lib/inputstream.py
@@ -119,14 +119,13 @@ class BufferedStream(object):
 
 def HTMLInputStream(source, encoding=None, parseMeta=True, chardet=True):
     if hasattr(source, "read"):
-        # Do no use .read(0) because of Python bug #20007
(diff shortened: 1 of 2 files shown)


gsnedders:
I'm so confused as to what's happened here. Why did this get dropped? Was I expecting you to update PR with that? I don't know. Looking over this again, `get_method_self` seems sanest, along with moving it within the following `if`. Closing this to replace it with #186.

gsnedders:
Oh, no, I'm being stupid. I'm thinking about JS semantics and not Python ones!

PR #137, opened 2014-01-05
Fix #124: Move to webencodings for decoding the input byte stream.

### LINKED ISSUE #124: Encoding Standard
html5lib has a bunch of [aliases for character encodings](https://github.com/html5lib/html5lib-python/blob/e269a2fd0aafcd83af7cf1e65bba65c0e5a2c18b/html5lib/constants.py#L2851). This should probably be aligned with the [Encoding Standard](http://encoding.spec.whatwg.org/).

I made a Python implementation, hopefully it will help: http://pythonhosted.org/webencodings/ (It mostly does aliases and BOM detection, the actual codec implementations are Python’s.)

Maybe related: #120

gsnedders:
Fixes #124. Note this currently fails a large number of tests because of <https://github.com/html5lib/html5lib-tests/issues/37>. As the changelog notes, this adds a dependency on @SimonSapin's webencodings.


gsnedders:
Build seems to be passing with XPASSes, hmm.

gsnedders:
@jgraham r?

PR #142, opened 2014-02-21
Fix arguments order in error message

sigmundch:


gsnedders:
@sigmundch  Given dicts have no order, how is there any issue here?

sigmundch:
Note that this swaps which key is associated with each value (e.g. `expectedName` should be `node.name` rather than `body`). So if you have:

```html
<div></span>
``` 

The error used to say:
`Unexpected end tag (div). Missing end tag (span).`

Now the error will say:
`Unexpected end tag (span). Missing end tag (div).`

gsnedders:
Oh. I can't read. Duh. :)

gsnedders:
rebased & landed as 3ebdd8bc2e4b751218e5189f8ff40e45d926efb2

PR #144, opened 2014-03-04
Fixing SVG close tag bug

sdcooke:
The test fails on current master - I'm new to this library so I may have fixed it in the wrong way or missed other places where it would fail but this solves my problem.


gsnedders:
See #221 instead of this, which fixes this as the spec does.

PR #145, opened 2014-04-04
Add useful message to the ParseError exception.

jdufresne:
Sometimes, I would like to be able to catch ParseError, and output its
message immediately.  Set the exception's message to the parser error
message.


gsnedders:
Rebased as 325aeaa. Thanks!

Issue #138, opened 2014-01-05
Add 'tel:' to sanitizer acceptable protocols?


igloox:
Most (if not all) mobile browsers can hyperlink telephone numbers, like so:

`<a href="tel:+44123456789">0123456789</a>``

When tapped / clicked, these numbers will open in the phone's dialler. Any objections to adding 'tel' to the list of acceptable_protocols in sanitizer.py?

willkg:
I just closed the relevant PR.

First, someone needs to research whether `tel:` is safe in all contexts. We'd want that for this to be added as a default. I think the answer is "no" based on this comment:

https://github.com/html5lib/html5lib-python/pull/149#issuecomment-97304684

Though that's conflating "valid" with "safe" a bit.

I don't think this is a good thing to have as a default. Given that, I'm going to close this out.

If someone can prove it's safe in all contexts, I'm game for re-opening and discussing it further.

gsnedders:
I'm pretty sure it is safe with the possible proviso of very old Android WebView, but I never sat down and figured this out. Validity is irrelevant for the sanitiser; we allow tons of invalid stuff.

PR #149, opened 2014-04-30
added 'tel' to list of acceptable hyperlink protocols

igloox:
As per my issue, #138

### DIFF
--- a/html5lib/sanitizer.py
+++ b/html5lib/sanitizer.py
@@ -138,7 +138,7 @@ class HTMLSanitizerMixin(object):
     acceptable_protocols = ['ed2k', 'ftp', 'http', 'https', 'irc',
                             'mailto', 'news', 'gopher', 'nntp', 'telnet', 'webcal',
                             'xmpp', 'callto', 'feed', 'urn', 'aim', 'rsync', 'tag',
-                            'ssh', 'sftp', 'rtsp', 'afs']
+                            'ssh', 'sftp', 'rtsp', 'afs', 'tel']
 
     # subclasses may define their own versions of these constants
     allowed_elements = acceptable_elements + mathml_elements + svg_elements


gsnedders:
In case anyone is curious about the status, this is mostly blocked on me wanting to check that nothing does anything crazy with anything like ``. Evidence either way appreciated.

agafonovdmitry:
Just to be sure, as per http://www.w3.org/TR/html5/embedded-content-0.html#attr-img-src :
> The src attribute must be present, and must contain a valid non-empty URL potentially surrounded by spaces referencing a non-interactive, optionally animated, image resource that is neither paged nor scripted.
So, I guess "tel" is NOT valid for image. As well as "mailto", by the way.

gsnedders:
It isn't *valid*, that's for sure. But the sanitizer doesn't currently try and maintain validity (because that's *a lot* more work), it's merely concerned with possible security consequences.

willkg:
This PR needs to be updated because the sanitizer has changed.

@agafonovdmitry, @dieselmachine: Can either of you update this PR?

willkg:
I'm going to close this out. It needs to be updated. Further, it's still not clear whether `tel:` is safe in all contexts. That needs to be researched. That's better done in the issue rather than in this PR.

willkg:
@gsnedders reopened #138. Given that, I'll reopen this and merge it.

willkg:
@igloox Do you have time to update this PR in the next week? If not, I can do it.

gsnedders:
@willkg unlikely

willkg:
This is out of date and the original author hasn't responded. We've got an issue for supporting it already. Given that, I'm going to close it out. If someone wants this change, they can write up a PR with the change and a corresponding test.

Issue #2, opened 2013-04-09
Support Jython


gsnedders:
From Google Code Issue 220:

Reported by steve@strassmann.com, Mar 1, 2013 
> What steps will reproduce the problem?
> Reproducible in Jython 2.5.2 and Jython 2.7b1
> ```
>>> import html5lib
import html5lib
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "lib/html5lib/__init__.py", line 14, in <module>
    from html5parser import HTMLParser, parse, parseFragment
  File "lib/html5lib/html5parser.py", line 33, in <module>
    import inputstream
UnicodeDecodeError: 'unicodeescape' codec can't decode bytes in position 48-54: illegal Unicode character
```


> What is the expected output? What do you see instead?
 jython cannot read inputstream.py.

> Please provide any additional information below.

> inputstream.py contains some seriously broken Unicode characters in the range 0xD800-0xDFFF, which are known as "unpaired surrogates".

> This has been closed as wont-fix: http://bugs.jython.org/issue1836

> It may be necessary to modify inputstream.py to not use these unicode character literals when running in Jython.

> n.b. a test for Jython:
>```
import platform
JYTHON = (platform.system() == 'Java')
```

 Apr 7 (2 days ago) geoffers 
> As I just commented on the Jython bug, I believe this is a bug in Jython not implemented Python as it is documented. I don't particularly want to add in hacks for a bug in Jython.

 Apr 7 (2 days ago) geoffers 
> Furthermore, if I do add a hack for it, we go into an infinite loop in the testsuite. I'm changing this bug to a more generic support Jython bug — no timeframe or even decision whether this is likely to happen.

gsnedders:
AND WE'RE DONE. AT LAST. Thanks to @jimbaker and @nonameentername, with #150, #168, and #185, we're finally here. Hopefully soon once pip picks up everything and virtualenv is running on Jython we can add Jython to tox and make it less likely for us to regress this.

PR #150, opened 2014-05-03

jimbaker:
Jython does not support isolated surrogates in unicode, including in unicode literals. This has been reported in https://github.com/html5lib/html5lib-python/issues/2 This bug is critical for Jython due to the fact that html5lib is a vendor lib for pip, and this is blocking pip from running on Jython.

For platforms besides Jython, this pull request allows for these surrogates to be constructed in literals, but through an additional step of indirection. For Jython itself, Jython's normal decode of literals will ensure that such invalid unicode strings cannot be constructed from any source.

To run this on Jython:

1. Install https://bitbucket.org/jimbaker/jython-socket-reboot, following these instructions: https://wiki.python.org/jython/JythonDeveloperGuide

2. Use this branch of pip to install nosetests, etc.: https://github.com/jimbaker/pip Note that tox is not yet supported - because we need to get pip running first! :)

Note that in the dev build, you will find executables in dist/bin, such as dist/bin/jython or dist/bin/pip

The jython-socket-reboot branch is nearly complete for merging into Jython; it is a major component of Jython 2.7.0 beta 3. (I'm a core dev of Jython.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Because we never want compiled Python
 __pycache__/
 *.pyc
+*.py$class
 
 # Ignore stuff produced by distutils
 /build/
--- a/html5lib/inputstream.py
+++ b/html5lib/inputstream.py
@@ -3,6 +3,7 @@ from six import text_type
 from six.moves import http_client
 
 import codecs
+import platform
 import re
 
 from .constants import EOF, spaceCharacters, asciiLetters, asciiUppercase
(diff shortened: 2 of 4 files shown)


tseaver:
I was hoping that the recent [jython 2.7b4 release](http://fwierzbicki.blogspot.com/2015/02/jython-27-beta4-released.html) somehow worked around this problem, but it is still borken.

tseaver:
Hmm, I don't know of any resolution here.  Are you in the process of merging #168?

gsnedders:
Just closing this because #168 is it's replacement rather than keeping both open.

gsnedders:
(We'll see if anything happens soon. Maybe.)

PR #154, opened 2014-05-19
Move the treewalker pretty printer to the treewalker module

gsnedders:
It previous lived merely in html5lib.tests.test_treewalker; it really is useful outside of it as a serialization of the treewalker. (Perhaps even as a `__repr__`?)


gsnedders:
Rebased/merged in commits to dee0321.

Issue #122, opened 2013-11-30
Warn on missing testdata submodule


SimonSapin:
Steps to reproduce:

```sh
git clone https://github.com/html5lib/html5lib-python.git
cd html5lib-python
tox
```

The tests fail with error messages like:
```
FileNotFoundError: [Errno 2] No such file or directory: '…/html5lib-python/html5lib/tests/testdata/encoding/chardet/test_big5.txt'
```

The reason is that `git clone` does not clone/checkout submodules by default. The fix is to run `git submodule update --init`.

When finding `html5lib/tests/testdata` empty, the test harness should print an appropriate warning with instructions.

PR #155, opened 2014-05-21
Fix issue #122: Add instructions to clone testdata submodule

blag:
Add instructions on how to get html5lib/tests/testdata/encoding/chardet/test_big5.txt from git.


gsnedders:
Sorry, I remember starting to review this and trying to work out how to deal with it in a better way in case test_encoding.py isn't run first (which I don't think it's guaranteed to be, depending on the order nose gets the files from the OS, which might be in inode order). At this stage, I don't think it's worth merging (sorry!), given so much of the test code is about to get rewritten with #182.

PR #156, opened 2014-05-22
Allow data URI schemes, with content type whitelist

blag:
Also, use proper URIs in the tests.

Note that the content type whitelist currently only contains 'image/png'; this can be filled out in a separate PR after a more complete discussion about allowable content types.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "2.7"
   - "3.2"
   - "3.3"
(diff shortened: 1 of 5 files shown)


gsnedders:
Bebased into f6448659094287650843ba61f9a1f81aa37743bc.

blag:
No problem, glad to see it made it in! :)

PR #157, opened 2014-05-22
Add setuptools support

blag:


gsnedders:
This seems to have snuck in part of e76c567c2.

PR #168, opened 2014-10-07

nonameentername:
Fixing flake8 trailing whitespace


gsnedders:
See #185.

PR #172, opened 2014-10-20
Allow unicode tag names and characters since that's what the tokenizer outputs

captbaritone:


gsnedders:
The filters should only be for treewalkers and not tokenizers; the tokenizer should be considered a private API.

gsnedders:
That said, it should *only* support unicode ones because the treewalker API requires unicode or a subclass.

gsnedders:
https://github.com/html5lib/html5lib-python/commit/40c3ba67f3a8573f64c3fec9a2aec9d7ab8b8f8d fixed basically all of this.

PR #174, opened 2014-11-24
Add expected failures

gsnedders:
This isn't quite finished: notably, we still have four failures in the tokeniser tests because we can't currently add the lone-surrogates tests because the expected failures file is UTF-8 and we can't have lone-surrogates there. Options are to have a WTF-8 codec or add some escapes to test data files (optionally just for new files, maybe?) or to just fix the tokeniser and worry about this later, I think. Suggestions?

### DIFF
--- a/html5lib/tests/expected-failures/tokenizer.dat
+++ b/html5lib/tests/expected-failures/tokenizer.dat
@@ -37,7 +37,4 @@
 <!DOCTYPE
 
 #data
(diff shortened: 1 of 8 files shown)


gsnedders:
This is basically abandoned in lieu of the expected_failures_pytest branch in my repo, given that provides a far better basis to be able to programmatically update expectations.

PR #181, opened 2015-02-09
Add some missing MathML elements and attributes.

fred-wang:
(in particular this adds the semantics/annotation/annotation-xml elements required to fix a very old planet bug http://www.maths-informatique-jeux.com/blog/frederic/?post/2013/01/22/Analysis-of-Lithium-s-algorithm#c142)


fred-wang:
Kurt McKee updated the list for feedparser. I'm adding encoding/other attribute to keep in sync. See https://github.com/distler/venus/commit/fdb788e798b58b92eec9f0a85eb53d7682f7fd8e#commitcomment-10724078

gsnedders:
*sighs*. Is there any sane way to programmatically generate a list of MathML elements/attributes? Though I guess that doesn't work because there still needs to be some review to make sure they're safe (for the sanitizer).

fred-wang:
Well, it you want to generate the list of MathML attributes/elements, you can use the official RelaxNG schema:

http://www.w3.org/Math/RelaxNG/

The HTML5 validator patches these schemas, but that should not change the list:

https://github.com/validator/validator/tree/master/schema/mml3

fred-wang:
@gsnedders This seems to have stalled... what's the plan here?

gsnedders:
@fred-wang Need to sit down and do a proper review of all the presentational MathML stuff to ensure there's nothing that could potentially break out of the inline box and take over the page.

fred-wang:
@gsnedders Here are some explanations if that can help to make progress. I would at least appreciate if the elements from 1) and 2) below and the associated attributes (notation and encoding) are added asap as they are used in real Web pages but unfortunately filtered out by Planet Venus. If you want, I can create a separate pull request to add them. Thanks.

1) semantics, annotation, annotation-xml: The first is a container to ask the renderer to display one "supported" child (similar to SVG switch element). The two others are just containers for simple text and arbitrary XML respectively. So all of this should be safe, although the annotation(-xml) elements can use the src attribute to refer to a separate resource and so you might want to sanitize the URL. See https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics.

2) menclose: This is just a container that adds enclosure or other symbols around an expression (e.g. circle or diagonal strike). See https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose

3) ms: Again, just a text container adding "quotes" around the text. See https://dxr.mozilla.org/mozilla-central/source/layout/mathml/mathml.css#51 for a simple CSS implementation.

4) mfenced: Again, just a container that adds fences and separators to its children. See https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mfenced

5) maligngroup, malignmark: Despite the unfortunate "malign" prefix, these are just empty elements that are used for alignment layout. Not natively supported in any Web rendering engines.

6) mglyph: This is used to add custom symbols as images. Essentially, equivalent to img so you'll have to handle sanitization of the "src" attribute the same way. Not natively supported in any Web rendering engines. See https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mglyph

7) mlabeledtr. Use in math tables to create a row with a label. Similar to a tr element with a "label" cell. Not natively supported in any Web rendering engines. https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mlabeledtr

8) mlongdiv, mscarries, mscarry, msgroup, msline, msrow, mstack: These are again layout elements (for "elementary math layout" like long division etc). Not natively supported in any Web rendering engines.

html5lib/sanitizer.py
```diff
@@ -24,11 +24,15 @@ class HTMLSanitizerMixin(object):
...
-                       'mprescripts', 'mroot', 'mrow', 'mspace', 'msqrt', 'mstyle', 'msub',
-                       'msubsup', 'msup', 'mtable', 'mtd', 'mtext', 'mtr', 'munder',
-                       'munderover', 'none']
+    mathml_elements = ['annotation', 'annotation-xml', 'maction', 'maligngroup',
+                       'malignmark', 'math', 'menclose', 'merror', 'mfenced',
+                       'mfrac', 'mglyph', 'mi', 'mlabeledtr', 'mlongdiv',
+                       'mmultiscripts', 'mn', 'mo', 'mover', 'mpadded',
+                       'mphantom', 'mprescripts', 'mroot', 'mrow', 'ms',
+                       'mscarries', 'mscarry', 'msgroup', 'msline', 'mstack',
+                       'mspace', 'msqrt', 'msrow', 'mstack' 'mstyle', 'msub',
```
fred-wang: mstack is added twice here :-(

fred-wang:
So I just published again a blog post whose layout is completely broken on planet Mozilla, planet WebKit and probably others relying on Planet Venus just because of the incomplete list in html5lib-python. I'm not sure why review is taking so long but please tell me how I can help.

broken-mathml

fred-wang:
External contributions do not seem welcome, so I'm closing this pull request and deleting my fork.

PR #184, opened 2015-04-17
Update html5parser.py

ta5ae:
Adding an initial definition of the new_phase variable as recently an application I work on managed to generate this error: a local variable 'new_phase' referenced before assignment.

### DIFF
--- a/html5lib/html5parser.py
+++ b/html5lib/html5parser.py
@@ -359,6 +359,7 @@ class HTMLParser(object):
     def resetInsertionMode(self):
         # The name of this method is mostly historical. (It's also used in the
         # specification.)
+        new_phase = None
         last = False
         newModes = {
             "select": "inSelect",


gsnedders:
Can you provide anything to reproduce the error? Almost certainly what you've written will merely delay an exception being thrown, it won't stop any exception from being thrown. To get such an exception there would imply the stack is empty, which would be very odd!

ta5ae:
Sorry for the late reply, I presumed the calling code had caused the error.  Here is a snapshot of the error log:

Local Vars

Variable | Value
-------- | -----
container | u'div'
stream | u'I keep others at a distance'
useChardet | True
encoding | None
self | &lt;html5lib.html5parser.HTMLParser object at 0x7fee33ec4810&gt;
innerHTML | True
parseMeta | True
kwargs | {}
...lib/python2.7/site-packages/html5lib/html5parser.py in reset
            self.resetInsertionMode()

gsnedders:
You wouldn't happen to have a stack trace? I'm pretty sure it should be impossible to have an empty stack, which is the only thing that should cause it.

ta5ae:
I hesitate to send a full stack trace due to the need to clean up a lot of private client / company data.

However, I believe the issue lies with us using your library in a threaded environment.  We no longer do.  This would account with the odd results.  Sorry for reporting the red herring.

gsnedders:
Ah, yes, that's plausible. We should better document that we're known to not be safe in a threaded environment. We should also fix it so we are safe.

Closing for now; if you happen to find evidence that it *isn't* down to threading, feel free to file a new issue and link to this.

PR #185, opened 2015-04-26

gsnedders:
This again. Replacement of #150 and #168 .


gsnedders:
This has landed in the form of b51828b20c3a3070342ef7e5f6d09676beb12303.

PR #186, opened 2015-04-26
Add a more general fix for #127 (CPy #20007) based on #136.

gsnedders:
This gets stuff working with `urllib.response.addurlinfo`. See #127 and #136.

### DIFF
--- a/html5lib/inputstream.py
+++ b/html5lib/inputstream.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, unicode_literals
+
 from six import text_type
-from six.moves import http_client
+from six.moves import http_client, urllib
 
 import codecs
 import re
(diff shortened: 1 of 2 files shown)


gsnedders:
@SimonSapin this may be of interest to you (maybe)

SimonSapin:
According to https://github.com/Kozea/WeasyPrint/commit/e14653c10260a7edd3b7416d6438b63939865336 , html5lib 0.999 already had a work around that was enough for WeasyPrint. Probably https://github.com/html5lib/html5lib-python/commit/18bc2787aa1156bd826eed62e54ec6bbb8403d76.

SimonSapin:
Oh, looking closer, this is a similar but different work around. I don’t remember WeasyPrint being affected by that one, though.

gsnedders:
Yeah, from memory you for were just trying to get it to be more complete, rather than needing it yourself. I think?

SimonSapin:
Maybe. But to be honest I’ve kind of moved on from Python :)

Issue #182, opened 2015-04-06
Move to py.test and take advantage of it


gsnedders:
We want to move to py.test so that:
- we can select to only run parser tests with one tree builder (and tree walker tests with only one tree walker, etc.)
- we can make use of the plugin API to create a means of dumping test failures and build support for expected failures into the testsuites from html5lib-tests
- it is easier to create nicer error messages
- better support for skipped and expected failures so we can walk towards having a constant number of tests across all platforms, optionally skipped

https://github.com/gsnedders/html5lib-python/tree/expected_failures_pytest is my current WIP branch implementing this.
### Needed to fix failures:
- [x] Make sure tests from generators run in a deterministic order
- [x] Use pytest-expect
### The rest:
- [ ] Move tokenizer and serialiser tests over to generating test objects from the JSON files.
- [ ] Refactor support.py.
- [x] Check the submodule is imported (see #122).

gsnedders:
I've, uh, been totally over-engineering this. Something along the lines of <https://github.com/dmerejkowsky/pytest-runfailed/blob/master/pytest_runfailed.py> would work fine.

gsnedders:
We've now landed the stuff most needed here. Yay!

willkg:
I think there are three things left to do here:

1. Update `html5lib/tests/README` to talk about pytest and not notes.
2. Move tokenizer and serialiser tests over to generating test objects from the JSON files.
3. Refactor support.py.

I can do the 1. I need more details on what 2 and 3 entail. Part of me thinks we should split 2 and 3 into separate issues.

@gsnedders Is that ok with you?

gsnedders:
@willkg lgtm

willkg:
I split 2 and 3 into separate issues and did a PR to remove that one bit about nose in the tests README. After that PR lands, I'll close this out.

willkg:
PR #366 landed and the other two things were spun off as separate issues, so everything outstanding is covered elsewhere now. Closing this out.

PR #198, opened 2015-07-21
Store expected failures and move to py.test

gsnedders:
This is #182. Also includes a fair bit of clean-up. Nevertheless, this is enough to get all tests passed/xfailing.


gsnedders:
Bah. The breakage is https://github.com/gsnedders/pytest-expect/issues/2.

gsnedders:
And having thought I'd fixed that, I've now run into a bug in pickle on Py3.2/3.3… <_<

gsnedders:
@jgraham I believe this is now in a position where it can land, with the history rewritten nicely. (Note GitHub's view appears to be sorted by author date, not by the actual parentage order…)

PR #202, opened 2015-08-09
Use a frozenset in sanitizer.py

mgilson:
Since all of those lists were only used for membership checking, this could provide a small speedup.

It also decreases the number of tests due to elements being repeated in some of the fields.


gsnedders:
#110 also fixes this. Though given that needs someone (read: me) to sit down and sort out the testsuite I guess we may as well merge this anyway. (Conflicts? Pff!)

mgilson:
Conflict resolution is why we use git, right?  :-).

gsnedders:
Bah, git doesn't cope that well when stuff is largely rewritten and moved to a different path. :)

PR #209, opened 2015-10-01
Support universal wheels

graingert:


gsnedders:
This is already done as part of #212 now.

PR #210, opened 2015-10-13
Fix HTMLSerializer on Python2 shouldn't assert unicode strings.

faerloev:
Currently the HTMLSerializer fails with an assert on unicode strings.


gsnedders:
Why? When does it fail? As far as I'm concerned, those assertions are valid. They should only be called with unicode strings as arguments. Where are they getting called with anything else?

gsnedders:
@faerloev do you have any traceback for the assertion failing?

PR #212, opened 2015-11-12
Move Dependency Into setup.py and Make Universal Wheel

### LINKED ISSUE #165: Wheel packaging format for pypi.org
Would it be possible to get a .whl universal package on Pypi for your package ?

If I'm right, it's just a matter of adding a setup.cfg file at root level with 
````
[wheel]
universal = 1
````
and then doing this to upload to pypi
````
python setup.py sdist bdist_wheel upload -r pypi
````

### LINKED ISSUE #179: Update documentation and requirements*.txt after 2.6 ordereddict change
As I commented in #177, the changelog (adding a dependency is a pretty major change!), readme (because it's now wrong!), and requirements*.txt files (because they don't require ordereddict on 2.6!) all need updating.

dstufft:
* Switches the setup.py to use setuptools.
* Automatically installs ``ordereddict`` on Python < 2.7.
* Adds extras which can be used to install optional dependencies like ``pip install html5lib[genshi,datrie]``. The ``lxml`` extra is conditional on CPython.
* Adds an ``all`` extra which will install all optional dependencies which correctly handles lxml on non CPython interpreters.
* Marks wheels as universal.

Fixes #165 
Fixes #179

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,9 @@ Change Log
 
 Released on XXX
 
-* XXX
+* Added ordereddict as a mandatory dependency on Python 2.6.
+* Added ``lxml``, ``genshi``, ``datrie``, ``charade``, and ``all`` extras that
+  will do the right thing based on the specific interpreter implementation.
 
 
 0.9999999/1.0b8
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal = 1
--- a/setup.py
+++ b/setup.py
@@ -58,11 +58,23 @@ setup(name='html5lib',
           'six',
       ],
       extras_require={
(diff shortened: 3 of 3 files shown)


dstufft:
Just an FYI:

* This will work as expected on setuptools >= 0.7 but I think it sill silently fail to install orderedict on 2.6 and lxml when using the lxml or all extras on older setuptools.
* This will work as expected on pip >= 6.0 but will silently fail to install ordereddict on 2.6 and lxml when using the lxml or all extras on older pip.

gsnedders:
Thanks!

gsnedders:
Bah, #179 still holds. The docs need to be updated to say we require ordereddict on 2.6 and require setuptools for install.

Ivoz:
@gsnedders you can use this syntax in requirements files these days:

    ordereddict; python_version < '2.7'

PR #214, opened 2015-11-23
Make lxml tree-builder coerce comments to work with lxml 3.5

gsnedders:
This enables the double-dash comment filter, and changes the filter to avoid trailing dashes which are equally illegal.


gsnedders:
0c551c9519e47f76f8f185089ed71cb9539b6e00

PR #216, opened 2015-12-02
Fix typos

jwilk:


gsnedders:
Thanks! Out of interest, did you use anything in particular to find these?

jwilk:
I used [mwic](http://jwilk.net/software/mwic).

PR #219, opened 2015-12-12
setup refactor v1.5

mindw:
* drop usage of charade - no longer maintained
* refactor setup for simplicity


gsnedders:
See #252.

PR #227, opened 2016-03-02
Fix python implementation marker for setuptools 2.20

paradoxxxzero:
Otherwise setup.py fail with `error in html5lib setup command: Invalid environment marker: python_implementation == 'CPython'`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,8 +65,8 @@ setup(name='html5lib',
 
           # A conditional extra will only install these items when the extra is
           # requested and the condition matches.
-          "datrie:python_implementation == 'CPython'": ["datrie"],
-          "lxml:python_implementation == 'CPython'": ["lxml"],
+          "datrie:platform_python_implementation == 'CPython'": ["datrie"],
+          "lxml:platform_python_implementation == 'CPython'": ["lxml"],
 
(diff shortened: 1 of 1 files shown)


gsnedders:
Bah, Travis failing because of pip dropping 3.2 support. On the upside, easy to fix: we drop 3.2 support.

gsnedders:
@paradoxxxzero Sorry for the delay; can you rebase this on master?

paradoxxxzero:
Done sorry for the delay :)

Chronial:
This breaks setup on older setuptools (this is my system setuptools):
```> python setup.py sdist
error in html5lib setup command: Invalid environment marker: platform_python_implementation == 'CPython'
> python --version
Python 2.7.6
> easy_install --version
setuptools 3.3
> cat /etc/issue
Ubuntu 14.04.3 LTS \n \l
```

gsnedders:
@Chronial see #230 for dealing with that; I presume in the future we'll break a relatively small number of setuptools releases.

gsnedders:
@Chronial see #231 for dealing with that; I presume in the future we'll break a relatively small number of setuptools releases.

Issue #10, opened 2013-04-09
Implement scripting disabled case.


gsnedders:
Since people have requested this.

coreybrett:
I'm having an issue with this as well. Would be nice to have it fixed.

michaelbrunnbauer:
We also need a fix for this.

PR #230, opened 2016-04-21
Implement inhead-noscript context, add script parameter to parse

neumond:
48 failing tests now pass. ```#script-on``` and ```#script-off``` parameters are now considered in testing.


neumond:
Looks like something unrelated started to selectively fail or pass depending on version of python.

html5lib/html5parser.py
```diff
@@ -22,18 +22,18 @@
 
 
 def parse(doc, treebuilder="etree", encoding=None,
-          namespaceHTMLElements=True):
+          namespaceHTMLElements=True, script=True):
```
gsnedders: Bikeshed moment: Can we call this `scripting`?

html5lib/html5parser.py
```diff
@@ -78,11 +78,12 @@ def __init__(self, tree=None, tokenizer=tokenizer.HTMLTokenizer,
...
                             getPhases(debug).items()])
 
-    def _parse(self, stream, innerHTML=False, container="div",
-               encoding=None, parseMeta=True, useChardet=True, **kwargs):
+    def _parse(self, stream, innerHTML=False, container="div", encoding=None,
+               parseMeta=True, useChardet=True, script=True, **kwargs):
 
         self.innerHTMLMode = innerHTML
         self.container = container
+        self.scriptMode = script
```
gsnedders: A real nit that doesn't really block anything, but given it really should be seconds to fix it: I'd prefer `scripting` here as well, given the spec calls it the "scripting flag".

html5lib/tests/tree_construction.py
```diff
@@ -51,14 +51,17 @@ def runtest(self):
         fragmentContainer = self.test['document-fragment']
         expected = self.test['document']
         expectedErrors = self.test['errors'].split("\n") if self.test['errors'] else []
+        script = True
+        if 'script-off' in self.test:
+            script = False
```
gsnedders: I wonder if this is a horrible suggestion, but I wonder if it's worthwhile doing something like `if 'script-on' in self.test and '<script' in self.test['data']: pytest.skip("scripting test")`?

html5lib/html5parser.py
neumond: Not a problem. I had doubts naming it.

gsnedders:
@neumond Thanks, this looks great! I've nothing but a few nits to complain about. :)

I'll try and have a look at why it's failing on Travis soon, FWIW.

(Also: this fixes #10.)

html5lib/html5parser.py
gsnedders: Also, I'd rather we default to False, given we don't actually implement scripting.

gsnedders:
@neumond: one hypothesis about the Travis failures: try recreating the expectation file on Python 2.7 (with optional extras installed).

html5lib/tests/tree_construction.py
neumond: This tests only ```<noscript>``` and the tag can be broken in tricky ways like ```!--noscript>```.

gsnedders: Right. But we're never going to pass anything containing `<script>document.write("<p>")</script>`, hence it makes sense to skip it. Script elements are only going to be in the tests if they mutate the DOM.

neumond: >tree-construction/scripted/webkit01.dat:1<script>document.write("2")</script>3
Huh. Is that supposed to pass? Tests under tree-construction/scripted/ contain neither ```#script-on``` nor ```#script-off```. No idea here.

neumond:
@gsnedders You were right, tests with cElementTree were not masked under py3. Thanks for solving pytest-expect py2/py3 issues in advance.

gsnedders:
@neumond I'll try and add a warning or error if you're updating the tests with the wrong config, given that's totally undocumented and even I'd forgotten what exactly was needed.

html5lib/tests/tree_construction.py
gsnedders: Uh, IIRC everything under `tree-construction/scripted/` should have `#script-on`. I'd expect `tree-construction/scripted/webkit01.dat` to fail currently—ideally we'd skip it because there's no way we can pass it without real scripting support.

gsnedders:
@neumond Oh, I hadn't noticed before, but: can you move the option to `HTMLParser.__init__` from `_parse`? Also, `test_treewalker.py` needs updated to enable/disable scripting.

gsnedders:
Why did I think that a good idea? No, `_parse` is the right place.

gsnedders:
Superseded by #243.

Issue #231, opened 2016-04-22
Invalid environment marker: platform_python_implementation


guettli:
I want to install html5lib editable and get this error:
```
host@host~$ pip install -e git+https://github.com/html5lib/html5lib-python.git#egg=html5lib
Obtaining html5lib from git+https://github.com/html5lib/html5lib-python.git#egg=html5lib
  Cloning https://github.com/html5lib/html5lib-python.git to ./src/html5lib
    Complete output from command python setup.py egg_info:
    error in html5lib setup command: Invalid environment marker: platform_python_implementation == 'CPython'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /home/foo_eins_d/src/html5lib
```

My pip version:

```
user@host:~$  pip --version
pip 7.1.2 from /home/foo_eins_d/local/lib/python2.7/site-packages (python 2.7)
```

Which version of pip is needed, or what else can I do to get `pip install -e ...` working?

gsnedders:
This is caused by a backwards incompatible change in setuptools. I can't find anything saying what version is needed quickly; sadly there's no way to be compatible with both old and new versions of setuptools. :(

I think we need `setuptools` 20.2?

gsnedders:
Perhaps https://github.com/pypa/setuptools/issues/503 means that the legacy syntax works with v20.6.6? In which case we can move back to the legacy syntax then we only don't support support 20.2 to 20.6.5, but otherwise works.

gsnedders:
https://github.com/pypa/packaging/issues/72#issuecomment-213826903 summarises this mess.

gsnedders:
Fixed by #234. We don't work with 20.2 to 20.6 (released mid-Feb till late-Mar 2016), but should work with both older and newer releases.

gsnedders:
Ergh. This is way worse than I thought. <https://gsnedders.github.io/python-marker-test/results.html> is a better summary of what supports what.

gsnedders:
So the options are:

 * `platform_python_implementation`: doesn't work before 18.5 (except on Jython/IronPython!)
 * `platform.python_implementation`: doesn't work before 20.6.6
 * `python_implementation`: doesn't work before 20.10.0 on Jython/IronPython, and doesn't work between 20.2 and 20.10.0 on other implementations.

[...]

PR #234, opened 2016-04-24
Use the platform.python_implementation because it's the most compat

gsnedders:
See <https://github.com/pypa/packaging/issues/72#issuecomment-213826903> for a discussion of the various setuptools different aliases support. This means we don't work with 20.2 to 20.6 (released mid-Feb till late-Mar 2016).


gsnedders:
Fixed #231.

PR #244, opened 2016-05-07
Allow self closing <col /> tags.

jdufresne:


jdufresne:
Any chance this could be reviewed and integrated? All feedback is welcome. I'm happy to work on the PR to meet any project standards.

gsnedders:
@jdufresne Sorry for both failing to respond to this initially and that comment!

In short, I'm trying to push out a release quickly with most of the API changes, and then trying to deal with parser changes after that. This rather got shoved into the latter category, for not really any good reason given it's a relatively small change…

CHANGES.rst
```diff
@@ -46,6 +46,8 @@ Released on XXX
 
 * **Drop support of charade, now that chardet is supported once more.**
 
+* Added support for self closing ``<col />`` tags.
```
gsnedders: In general, we keep small changes out of the changes file. (Does the contributing doc say that? If not, bah! That really should've been caught before now…)

html5lib/html5parser.py
```diff
@@ -1706,6 +1706,7 @@ def startTagColgroup(self, token):
 
         def startTagCol(self, token):
             self.startTagColgroup(impliedTagToken("colgroup", "StartTag"))
+            token["selfClosingAcknowledged"] = True
```
gsnedders: Do we actually need this here before it gets returned and reprocessed? I'd like to think that the reprocessing handled this fine. (If it doesn't, it should be fixed!)

gsnedders: Oh, it doesn't. Hopefully that's just changing `token` to `new_token` in the loop.

CHANGES.rst
jdufresne: Removed from changes.

html5lib/html5parser.py
jdufresne: Sorry, I'm not quite sure what you're suggesting I change here.

I based this change on other self closing tag examples. Such as:

* [`link`](https://github.com/html5lib/html5lib-python/blob/fbce67d1c3f79835b1f9eca172a1171a9df9acc0/html5lib/html5parser.py#L683-L686)
* [`meta](https://github.com/html5lib/html5lib-python/blob/fbce67d1c3f79835b1f9eca172a1171a9df9acc0/html5lib/html5parser.py#L688-L691)
* [`hr`](https://github.com/html5lib/html5lib-python/blob/fbce67d1c3f79835b1f9eca172a1171a9df9acc0/html5lib/html5parser.py#L1179-L1185)

Is `col` a different case or are you suggesting a fundamental change to all of these?

Which loop are you referring to, `mainLoop`? In this loop `new_token` is `None` by the time it checks for the `non-void-element-with-trailing-solidus` error. So I'm a bit confused.

gsnedders: So some things return tokens to reprocess them in another state, such as for `col` in the "in table" insertion mode. Seemingly, none of the `token["selfClosingAcknowledged"]` assignments we have currently happen anywhere where a token is returned.

However, given the spec only defines the `col` start tag token as having it acknowledged in only one place (the "in column group" insertion mode), we should ideally do likewise, and the reprocessing of tokens in the `mainLoop` should handle this. This means we somehow need to keep track of the previous value of `new_token` given as you point out it's always `None` after the loop (you can tell I didn't look closely!), presumably by adding a `prev_token` variable or some such.

gsnedders: (I hope that clarifies what I meant, if not, let me know!)

jdufresne: That helps a lot! Thanks. I'll see what I can do.

jdufresne:
Thanks for the guidance, that change worked! All additional feedback welcome.

gsnedders:
Awesome, at quick glance that looks fine, will take a closer look later. Thanks!

Issue #119, opened 2013-11-02
Various html5lib improvements


kovidgoyal:
This is just a FYI, I have been working on a modified version of html5lib that achieves the following goals:

1) Preserves attribute order
2) Optionally includes line and column number information when parsing
3) Handles XML namespaces correctly, so that if you happen to parse an XHTML document with html5lib you dont lose all the namespace information
4) Create a new treebuilder for lxml
5) Various performance improvements

Using my new lxml treebuilder parsing performance with line numbers and attribute order preservation is the same as for vanilla html5lib with its builtin treebuilder. The speed improvements come mainly from the new lxml builder and an optimized inputstream class for in memory streams.

I make no claims as to the relevance of my work for html5lib. I am just sharing it with you as a way of giving back. You are welcome to use the patches or not. Feel free to ask if you need any clarification.

The code is in https://github.com/kovidgoyal/calibre/tree/master/src/html5lib (these are the changes to html5lib itself)

and the lxml builder is in:
https://github.com/kovidgoyal/calibre/blob/master/src/calibre/ebooks/oeb/polish/parsing.py

scoder:
> Element.attrib (which generates a python dictionary from the underlying libxml2 structure).

No, it actually uses its own mapping class based on the underlying XML tree data. The order is therefore preserved.

> lxml, FWIW, explicitly documents everywhere that the order is "arbitrary".

"Arbitrary" in the sense that it depends on where the attributes came from. If from the parser, then it's document order. If added later, it might make a difference whether the attribute already existed before, but AFAIR it still gets appended at the end. It's up to libxml2, and lxml doesn't give any guarantees here on its own. It's unlikely that the current behaviour of libxml2 is ever going to change, though, as it is considered a feature.

Issue #97, opened 2013-07-22
Add element.sourceline (lxml)


SimonSapin:
Elements in lxml have a [`sourceline` attribute](http://lxml.de/api/lxml.etree._Element-class.html#sourceline), but it is always `None` in trees parsed with html5lib.

html5lib also appears to [keep track of source line numbers for error messages](https://github.com/html5lib/html5lib-python/blob/da9459fd2fb172c8725f27fa75acca7a601de368/html5lib/html5parser.py#L246).

I’d like to connect the two so that WeasyPrint can do its own error reporting based on line numbers of HTML elements. It seems like the attribute should be set [in etree_lxml.py](https://github.com/html5lib/html5lib-python/blob/da9459fd2fb172c8725f27fa75acca7a601de368/html5lib/treebuilders/etree_lxml.py#L215), but I don’t know how to access the input stream from there.

Suggestions?

Issue #87, opened 2013-07-06
Add line/column number in tokens from HTMLTokenizer


Naddiseo:
Would it be possible to add line and column tracking to the tokens returned from HTMLTokenizer?

PR #245, opened 2016-05-07
Selected patches from Calibre

gsnedders:
See #119. CC @kovidgoyal.

This cherry-picks a few things from <https://github.com/gsnedders/html5lib-python/commits/calibre-patches>, which was a complete set of Calibre's patches from November 2013. <https://github.com/kovidgoyal/calibre/commits/master/src/html5lib> has very little changed in it since then, primarily a move to 0.999999-dev and a separate downstream fix for https://github.com/html5lib/html5lib-python/commit/0c551c9519e47f76f8f185089ed71cb9539b6e00.

So, of those on that branch…

 * <https://github.com/gsnedders/html5lib-python/commit/49f37d2724b117b1eaa1ba7a23d27aef14db6932>, https://github.com/gsnedders/html5lib-python/commit/64e8b0bd8e9bc4dbb8339ab0e094ef933035339c, and https://github.com/gsnedders/html5lib-python/commit/cc9f28af4859663738c67971f9893e2558f86138 are all things I'm against landing upstream (they should just be special-cased in the downstream tree builder rather than requiring invoking separate handling for all tree builders).
 * https://github.com/gsnedders/html5lib-python/commit/7702d80e52c1d1c90ea41af6808623c78b518f01 is something I find very inelegant, and goes against the recommendations laid out in PEP 8. Also, with CPython, in the `True`/`False` cases it's likely slower, therefore failing at its stated goal, as it results in more byte code and `POP_JUMP_IF_FALSE` and `POP_JUMP_IF_TRUE` special-case the condition being `True` or `False` (oddly, they don't specialise `None`, though it is in `PyObject_IsTrue`; if that makes any notable performance difference then I'd suggest fixing that in CPython).
 * <https://github.com/gsnedders/html5lib-python/commit/a2d2e05fb667683db9e020e7a2629e62fdc95832> is unlikely to land because the input stream isn't public API (not that we document what is anywhere…).
 * <https://github.com/gsnedders/html5lib-python/commit/1576515fcd62af2969e387c92846790ec177a6db> in principle is something we want, but I think we want for all tokens and not just elements. (See also #87 and #97.) That or something similar will likely land in future.
 * <https://github.com/gsnedders/html5lib-python/commit/124e97565affa33dd36fd29c874020c72cce29fb> is included here, after <https://github.com/html5lib/html5lib-python/commit/9dca7d8e8e2eb95f8e0f9acba0a29386129b272e> which fixes the DOM tree builder to follow the tree builder API correctly. (It exposed bugs in the DOM tree builder, yay!)
 * <https://github.com/gsnedders/html5lib-python/commit/db0b5a02a2cb56d0410b878005ca5fdfa91961b2> is what most of the commits on this PR deal with, fixing up small issues with the Calibre code.


gsnedders:
This seems to have broken 2.6 badly. Huh.

gsnedders:
Oh, right, this is `viewkeys` (and dictionary views in general) not existing on 2.6.

kovidgoyal:
How do you suggest I override the application of attributes for html and body tags in my builder? Since without those patches, it would require overriding the entire getPhases() method in html5parser.py Remember that the problem those patches is solving is that there can be multiple <html> and <body> tags in malformed documents that I need to handle by merging the attributes from those tags into the first <html>/<body> tag. Doing that after parsing would require an whole extra tree traversal which can be slow for large trees. 

If you dont want to merge gsnedders@a2d2e05 then how do you suggest I replace the the stream input class? The one in html5lib is too slow. The only alternative I can see is monkey patching -- which is less than optimal for obvious reasons.

gsnedders:
@kovidgoyal I'll take a look at dealing with html/body attributes later (I'm literally amount to board a plane).

When it comes to the input stream, if it yields good perf increases when given a byte/unicode object we should just specialise them in html5lib.

kovidgoyal:
Sure you are welcome to take the input stream class from calibre for dealing with unicode objects. It is faster because it avoids wrapping the unicode in StringIO. And it actually implements tracking of positions. For my use case, that is important, since I need line and col numbers.

PR #250, opened 2016-05-10
Update html5lib-tests data

gsnedders:


gsnedders:
We should wait till https://github.com/html5lib/html5lib-tests/pull/76 and https://github.com/html5lib/html5lib-tests/pull/70 land.

PR #263, opened 2016-06-06
Try dealing with the backwards incompatible changes in setuptools again

gsnedders:
This tries to address #231 again, reverting to platform_python_implementation. This should work everywhere with setuptools 18.5 above; see #231 for more discussion of this.

### DIFF
--- a/requirements-install.sh
+++ b/requirements-install.sh
@@ -6,7 +6,7 @@ if [[ $USE_OPTIONAL != "true" && $USE_OPTIONAL != "false" ]]; then
 fi
 
 # Make sure we're running setuptools >= 18.5
-pip install -U pip setuptools
+pip install -U pip setuptools>=18.5
 
 pip install -U -r requirements-test.txt
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 six
 webencodings
 ordereddict ; python_version < '2.7'
+setuptools>=18.5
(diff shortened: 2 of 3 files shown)


gsnedders:
Hmm, maybe shouldn't run with loads of setuptools versions…

jraby:
For the record, checking version of setuptools in the body of `setup.py` in this manner makes it impossible to install/upgrade setuptools at the same time as html5lib.

For example this fails:

```
cat >req.txt <<EOF
setuptools>=18.5
html5lib
EOF

virtualenv testenv
testenv/bin/pip install -r req.txt

Collecting setuptools>=18.5 (from -r req.txt (line 1))
  Using cached setuptools-24.0.3-py2.py3-none-any.whl
Collecting html5lib (from -r req.txt (line 2))
  Using cached html5lib-0.999999999.tar.gz
    Complete output from command python setup.py egg_info:
    html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 18.2)

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/49/p3xmv77x6fnb2lwl5kpvmvs40000gp/T/pip-build-0vYISn/html5lib
You are using pip version 7.1.2, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

gsnedders:
@jraby can you file a new issue for that? that's a good point, and should check if removing that `if` makes it work (I assume it does!).

PR #265, opened 2016-06-08
A bunch of testing fixes

### LINKED ISSUE #122: Warn on missing testdata submodule

### LINKED ISSUE #237: Add error if update-xfail is used in wrong environment
Need to make sure options are installed and `ElementTree`/`cElementTree` divergence.

gsnedders:
Fixes #122, fixes #237.


gsnedders:
Pretty sure this can't be done with better coverage.

PR #272, opened 2016-07-13
Get rid of getPhases

gsnedders:
With this we no longer include "process the token using the rules for" phases in the debug log.

This also needs perf review given it touches `mainLoop`.

(If you want to view the diff, append `?w=1` to the URL to ignore whitespace, otherwise the reindenting of the phases just dominates.)

### DIFF
--- a/html5lib/html5parser.py
+++ b/html5lib/html5parser.py
@@ -1,7 +1,5 @@
 from __future__ import absolute_import, division, unicode_literals
-from six import with_metaclass, viewkeys
-
-import types
+from six import viewkeys
 
 from . import _inputstream
 from . import _tokenizer
@@ -71,18 +69,6 @@ def parseFragment(doc, container="div", treebuilder="etree", namespaceHTMLElemen
     return p.parseFragment(doc, container=container, **kwargs)
 
 
-def method_decorator_metaclass(function):
-    class Decorated(type):
-        def __new__(meta, classname, bases, classDict):
-            for attributeName, attribute in classDict.items():
-                if isinstance(attribute, types.FunctionType):
-                    attribute = function(attribute)
-
-                classDict[attributeName] = attribute
-            return type.__new__(meta, classname, bases, classDict)
-    return Decorated
-
-
 class HTMLParser(object):
     """HTML parser
 
@@ -112,6 +98,7 @@ class HTMLParser(object):
 
         # Raise an exception on the first error encountered
         self.strict = strict
+        self.debug = debug
 
         if tree is None:
             tree = treebuilders.getTreeBuilder("etree")
@@ -119,7 +106,7 @@ class HTMLParser(object):
         self.errors = []
 
         self.phases = {name: cls(self, self.tree) for name, cls in
-                       getPhases(debug).items()}
+                       _phases.items()}
 
     def _parse(self, stream, innerHTML=False, container="div", scripting=False, **kwargs):
 
@@ -201,6 +188,9 @@ class HTMLParser(object):
         DoctypeToken = tokenTypes["Doctype"]
         ParseErrorToken = tokenTypes["ParseError"]
 
+        type_names = {value: key for key, value in tokenTypes.items()}
+        debug = self.debug
+
         for token in self.tokenizer:
             prev_token = None
             new_token = token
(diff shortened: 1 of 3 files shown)


willkg:
@gsnedders Is this important? There's no explanatory issue, so I'm not sure what the value is here and whether we should spend time on it for 1.0 or not.

gsnedders:
@willkg Reduction in complexity and potentially performance gains by reducing indirection.

willkg:
Ok. I'm going to push this off until after 1.0, then.

html5lib/html5parser.py
```diff
@@ -201,6 +188,9 @@ def mainLoop(self):
         DoctypeToken = tokenTypes["Doctype"]
         ParseErrorToken = tokenTypes["ParseError"]
 
+        type_names = {value: key for key, value in tokenTypes.items()}
+        debug = self.debug
```
jayaddison, line 192: @gsnedders Is there a reason to bring `tokenTypes` and `self.debug` into local method variables here?

At [L225](https://github.com/html5lib/html5lib-python/pull/272/files#diff-88cdf3a2fd8f807dae741251459c88e1e10ed54666f13ab7fdc6f4d47ae08648R225) we could compare using `if self.debug` and similarly at [L226](https://github.com/html5lib/html5lib-python/pull/272/files#diff-88cdf3a2fd8f807dae741251459c88e1e10ed54666f13ab7fdc6f4d47ae08648R226) we could use `info = {"type": tokenTypes[type]}`.

(am guessing it's possibly an artifact of some IDE-based refactoring?  Hopefully a straightforward cleanup either way)

html5lib/html5parser.py
```diff
@@ -393,2386 +394,2380 @@ def parseRCDataRawtext(self, token, contentType):
...
+        # Note the caching is done here rather than BoundMethodDispatcher as doing it there
+        # requires a circular reference to the Phase, and this ends up with a significant
+        # (CPython 2.7, 3.8) GC cost when parsing many short inputs
+        name = token["name"]
+        # In Py2, using `in` is quicker in general than try/except KeyError
+        # In Py3, `in` is quicker when there are few cache hits (typically short inputs)
+        if name in self.__startTagCache:
+            func = self.__startTagCache[name]
         else:
-            return type
```
jayaddison, line 423: Nice :)  I get some minor worries about difficult-to-debug future issues when built-in Python keywords like `type` get used as variables / returned as values.  Removing this is a nice improvement :+1:

html5lib/tests/test_parser2.py
```diff
@@ -68,7 +68,6 @@ def test_debug_log():
                 ('dataState', 'InBodyPhase', 'InBodyPhase', 'processStartTag', {'name': 'p', 'type': 'StartTag'}),
                 ('dataState', 'InBodyPhase', 'InBodyPhase', 'processCharacters', {'type': 'Characters'}),
                 ('dataState', 'InBodyPhase', 'InBodyPhase', 'processStartTag', {'name': 'script', 'type': 'StartTag'}),
-                ('dataState', 'InBodyPhase', 'InHeadPhase', 'processStartTag', {'name': 'script', 'type': 'StartTag'}),
```
jayaddison, line 71: Good catch!  How did you discover this duplicate entry, out of interest?

html5lib/html5parser.py
```diff
@@ -393,2386 +394,2380 @@ def parseRCDataRawtext(self, token, contentType):
...
+    startTagHandler = _utils.MethodDispatcher([
+        ("html", startTagHtml),
+        ("noframes", startTagNoFrames)
+    ])
+    startTagHandler.default = startTagOther
+
+# pylint:enable=unused-argument
+
+
+_phases = {
```
jayaddison, line 2745: The core of the improvement - looks good generally.  What do you think about going one step further and moving this to the [initialization of the `HTMLParser`'s self.phases at L121](https://github.com/html5lib/html5lib-python/pull/272/files?w=1#diff-88cdf3a2fd8f807dae741251459c88e1e10ed54666f13ab7fdc6f4d47ae08648L121)?

jayaddison (commented):
NB: I've no write or commit access here, but to me these changes look good.

From local checkout & review:

- [x] Tests pass
- [x] Performance impact (for `parser` benchmarks, since that's what this code affects): neutral (cpython 3.9.1)

Although there doesn't seem to be a performance improvement from this, in my opinion it does work towards making the code simpler and opens up future refactoring and cleanup work that _is_ likely to improve performance in more significant ways.

ambv:
This was merged as #567.

PR #274, opened 2016-07-14
0.99999999 release! Let's party!

gsnedders:
So almost 1.0. So almost…


crgwbr:
Just a suggestion: https://www.python.org/dev/peps/pep-0440/#pre-releases.

PR #278, opened 2016-07-15
Add a note about API stability and major breakages

gsnedders:
cc @nabeelio


nabeelio:
:metal:

willkg:
This is pretty outdated. For one thing, Bleach supports recent versions now. I'm going to close this out. If there's an underlying issue that still needs to be fixed, then someone can open up an issue and we can figure out what to do there.

PR #284, opened 2016-07-22
Remove setuptools check from setup.py

### LINKED ISSUE #280: Installing html5lib in a fresh venv fails if setuptools is not up to date
Since #263 has been merged, it is not possible to install/upgrade setuptools at the same time as html5lib.

For example this fails:

```
cat >req.txt <<EOF
setuptools>=18.5
html5lib
EOF

virtualenv testenv
testenv/bin/pip install -r req.txt

Collecting setuptools>=18.5 (from -r req.txt (line 1))
  Using cached setuptools-24.0.3-py2.py3-none-any.whl
Collecting html5lib (from -r req.txt (line 2))
  Using cached html5lib-0.999999999.tar.gz
    Complete output from command python setup.py egg_info:
    html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 18.2)

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/49/p3xmv77x6fnb2lwl5kpvmvs40000gp/T/pip-build-0vYISn/html5lib
You are using pip version 7.1.2, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

jdufresne:
Fixes #280

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,16 +2,10 @@ from __future__ import print_function
 
 import ast
 import codecs
-import sys
 
 from os.path import join, dirname
-from setuptools import setup, find_packages, __version__ as setuptools_version
-from pkg_resources import parse_version
+from setuptools import setup, find_packages
 
-if parse_version(setuptools_version) < parse_version("18.5"):
-    print("html5lib requires setuptools version 18.5 or above; "
-          "please upgrade before installing (you have %s)" % setuptools_version)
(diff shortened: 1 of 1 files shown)


gsnedders:
Having landed #285 we've fixed this as well as support for older setuptools releases.

PR #285, opened 2016-07-22
Monkeypatch pkg_resources to always use _markerlib

jayvdb:
Prior to setuptools 20.10.0 there is patchy support for environment markers,
and setup.py fails while parsing them.
html5lib requires at least setuptools 18.5 for its environment markers.

However, @gsnedders developed a way to monkey patch pkg_resources
so that it always uses _markerlib, which allows all environment markers
to be used for any version of setuptools.  Some patching of _markerlib
is also required so that it works on Python 3 also.

On removing the dependency for setuptools>=18.5, pip partially fails on
Python 2.6 with an error `Double requirement given: ordereddict`
and does not install the requirements.txt.
Fixed by removing ordereddict from requirements-test.txt

### DIFF
--- a/requirements-install.sh
+++ b/requirements-install.sh
@@ -5,9 +5,6 @@ if [[ $USE_OPTIONAL != "true" && $USE_OPTIONAL != "false" ]]; then
   exit 1
 fi
 
-# Make sure we're running setuptools >= 18.5
-pip install -U pip setuptools>=18.5
-
 pip install -U -r requirements-test.txt
 
 if [[ $USE_OPTIONAL == "true" ]]; then
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -4,4 +4,3 @@ flake8
 pytest
 pytest-expect>=1.1,<2.0
 mock
-ordereddict ; python_version < '2.7'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 six
 webencodings
 ordereddict ; python_version < '2.7'
-setuptools>=18.5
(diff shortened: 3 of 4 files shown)


jayvdb:
Also fixes #280 as an alternative to #284

jayvdb:
@gsnedders and I had a discussion about this general problem at https://github.com/pypa/packaging/issues/72

The long and short of it is that it works on a very large array of Python & setuptools versions (https://travis-ci.org/jayvdb/tox-travis/builds/145144692)

gsnedders:
Hey, thanks for this. Can you add a (temporary) commit that tests (on Travis) loads of different pip and setuptools versions? (Let's try down to setuptools 3.3 and pip 6.0?)

I'd much rather test this with all of our relatively complex usage of markers directly.

jayvdb:
Sure.  Are you happy with the selection of versions being roughly the latest patch release of each minor release?

(that is what I did with https://travis-ci.org/jayvdb/tox-travis/builds/145144692 )
I'll start setting that up, and rather randomly mix setuptools and pip versions.

Do you want the commit to be on this PR, so that the record is part of this projects travis builds?

jayvdb:
no problems with normal setuptools use; pip 6.0 is definitely the minimum for the environment markers, but otherwise works ok.  Doing a large matrix build now with and without optional.  Will be a while before they are done.
https://travis-ci.org/jayvdb/html5lib-python/builds/147372074
https://travis-ci.org/jayvdb/html5lib-python/builds/147372330

gsnedders:
Awesome, thanks!

PR #287, opened 2016-07-22
Allows uppercase color codes in style

komaldembla:
Adds uppercase check (A-F) in the css regex to pass css of the following format:
 border-top: 2px #DA4534 solid;

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2013 James Graham and other contributors
+Copyright (c) 2006-2013 James Graham, Google Inc. and other contributors
 
 Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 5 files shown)


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2013 James Graham and other contributors
+Copyright (c) 2006-2013 James Graham, Google Inc. and other contributors
```
komaldembla: Updated the file as suggested.

willkg:
Sorry that no one has gotten to this, yet! This needs to be updated to reflect the sanitizer changes.

@komaldembla Can you update this PR? If not, I can look into it later.

willkg:
This is way out of date and the original author hasn't responded. Given that, I opened up issue #354 and I'm going to close this out.

PR #291, opened 2016-07-26
Require flake8 to be < 3.0 for Python 2.6 support

### LINKED ISSUE #290: flake8 v3 does not support Python 2.6
flake8 v2 supported Python 2.6, and is used by the Travis builds.

flake8 v3 has come out of beta, though is full of bugs, but more importantly it comes with a clear intent to not support Python 2.6, which is the trendy thing to do.
https://gitlab.com/pycqa/flake8/commit/cde783d135953d7d16a1373f13b066731083e477
https://gitlab.com/pycqa/flake8/commit/c8ad626525631314decbb5bd544be47102610fd0

To support continued linting of Python 2.6, raise upstream and/or pin flake8<3 at least on Python 2.6

Or drop support for Python 2.6.  Is that an option?  If not, it would be good to clearly explain why Python 2.6 support is needed, here and upstream.

gsnedders:
Fixes #290.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,6 +1,6 @@
 -r requirements.txt
 
-flake8
+flake8<3.0
 pytest
 pytest-expect>=1.1,<2.0
 mock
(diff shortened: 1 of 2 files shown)


jgraham:
r+

jayvdb:
👍 :(

PR #294, opened 2016-07-28
Fix spider

jayvdb:
This change is [Reviewable](https://reviewable.io/reviews/html5lib/html5lib-python/294)

### DIFF
--- a/utils/spider.py
+++ b/utils/spider.py
@@ -112,6 +112,13 @@ class Spider(object):
             newUrls.add(urllib_parse.urlunsplit(splitURL))
         urls = newUrls
 
+        toVisit = self.check_robots(urls)
+        toVisit = self.check_headers(toVisit)
+
+        self.visitedURLs.update(urls)
+        self.unvisitedURLs.update(toVisit)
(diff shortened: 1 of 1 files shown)


jayvdb:
We could probably replace most of this with an existing library.

Ms2ger:
Can we use `six` for the `urllib` imports?

jayvdb:
Using `six` here looks good; I checked and it provides `robotparser`, since 2013 (https://bitbucket.org/gutworth/six/pull-requests/5/create-sixmovesurllib/diff / https://bitbucket.org/gutworth/six/commits/1f2c7f5d14be9027d180aa00138a1d29c8f48a6f).
Should I find a minimum version for six support?  Currently it doesnt have one.

gsnedders:
As far as I'm aware, nobody's actually run it for years hence it being badly broken. We should probably just kill it at this point, as it likely has no real use.

willkg:
Given the previous comment, I'm going to close this out. I opened #349 to cover removing `utils/spider.py`.

PR #299, opened 2016-09-07
Allow disabling the stripping of HTML comments in sanitizer

jdufresne:
Can be used by libraries like
[bleach](https://github.com/mozilla/bleach) to upgrade to the new
sanitizer API.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Change Log
 ----------
 
+UPCOMING
+~~~~~~~~
+
+* Added ``strip_comments`` parameter to sanitizer. When ``False`` HTML comments
+  will not be stripped.
+
 0.999999999/1.0b10
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


gsnedders:
Do we actually want to allow this? This has security ramifications, given IE's conditional comments and I want to say some of the things that end up being parsed as bogus comments equally do. (That said, `allowed_elements` etc. also have security ramifications.)

gsnedders:
@jdufresne do I take you closing this to mean that we don't actually want this and it doesn't block bleach?

jdufresne:
@gsnedders I am not a maintainer of bleach,

TBH, I'm not interested in actually using the `strip_comments` functionality of this PR. In my projects I want the comments stripped.

A while back, I was interested in using bleach for a project. However, its `setup.py` required an older version of html5lib. Older than I was already using. I investigated why. A bleach unit test using the old `strip_comments` was failing with the newer html5lib. So I thought it would be progress towards submitting a PR to the bleach team with regards to updating the html5lib dependency.

I have since dropped the idea and built a minimal HTML cleanser using the new html5lib code. So for my project, it already meets my needs.

As I don't intend to follow up with bleach or use the `strip_comments` I have no need for it. Please follow through with how you see best.

Issue #298, opened 2016-08-31
Older versions of six may be incompatible


keithcallenberg:
Please disregard this if there are any overriding implications of requiring an updated version of six, but I found I needed an upgraded version in order to import html5lib.

With six==1.5.2 I receive an ImportError:

```
File "/usr/local/lib/python2.7/dist-packages/weasyprint/__init__.py", line 19, in <module>
    import html5lib  # noqa
  File "/usr/local/lib/python2.7/dist-packages/html5lib/__init__.py", line 16, in <module>
    from .html5parser import HTMLParser, parse, parseFragment
  File "/usr/local/lib/python2.7/dist-packages/html5lib/html5parser.py", line 2, in <module>
    from six import with_metaclass, viewkeys, PY3
ImportError: cannot import name viewkeys
```

This error goes away when I add six=1.10.0 to my requirements.

jayvdb:
The travis build errors on #301 also show that `html5lib/html5parser.py` also imports `six.viewkeys`, which was introduced in six 1.9.0 .

a137d14b by @kovidgoyal in 2013 is where `six.viewkeys` was introduced.  It is very easy to avoid this, as that commit also included a fallback.

gsnedders:
I think I decided that 1.9 being over a year old was a good enough reason to rely on it, as if you're using a recent version of html5lib, you're obviously not opposed to running recent versions of software. Aside from pip, setuptools, and wheel we have no reason not to rely on recent versions of dependencies, as far as I'm aware.

Obviously, if you disagree, say!

mdxs:
By silence procedure and #301 being merged, I would say this can be closed now.

madelinezec:
I'm having this exact error, but I have six==1.10.0 installed. Any other reason this could be happening?

PR #301, opened 2016-09-15
Declare implicit dependency on Six 1.9 or higher

amorde:
This project uses a syntax for importing urllib.parse from six which was introduced in Six version 1.4

The import can be found [here](https://github.com/html5lib/html5lib-python/blob/master/html5lib/filters/sanitizer.py#L6)

The relevant change in Six can be found [here](https://bitbucket.org/gutworth/six/commits/1f2c7f5d14be9027d180aa00138a1d29c8f48a6f?at=1.4.0#Lsix.pyT188)

This can cause an issue if a project has a previous version of Six (say version 1.3) installed when installing html5lib

EDIT:
importing `viewkeys` in html5parser.py requires six 1.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ cache:
 env:
   - USE_OPTIONAL=true
   - USE_OPTIONAL=false
+  - SIX_VERSION=1.9 USE_OPTIONAL=true
 
 install:
   - bash requirements-install.sh
--- a/requirements-install.sh
+++ b/requirements-install.sh
@@ -11,6 +11,10 @@ if [[ $USE_OPTIONAL == "true" ]]; then
   pip install -U -r requirements-optional.txt
 fi
 
+if [[ $SIX_VERSION != "false" ]]; then
+  pip install six==$SIX_VERSION
+fi
+
 if [[ $CI == "true" ]]; then
   pip install -U codecov
 fi
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-six
+six>=1.9
 webencodings
 ordereddict ; python_version < '2.7'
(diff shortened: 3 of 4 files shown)


jayvdb:
I need this for #294.

jayvdb:
It would be nice to have a test job that explicitly requires and uses six 1.4 , so that newer features are not inadvertently used without upgrading the requirement.

amorde:
@jayvdb That sounds like a great idea, I'm not really sure how to go about doing that though.

jayvdb:
You could add a job with

```diff
 env:
   - USE_OPTIONAL=true
   - USE_OPTIONAL=false
+  - USE_SIX_14=true
```

And then in `requirements-install.sh` look for that flag and force install 1.4 (and if anything upgrades it to a more recent version, fix it).

amorde:
@jayvdb It appears the minimum required version is actually **1.9**. Good call on the testing thing.

I'll squash these commits and do a force push

jayvdb:
In addition to the Travis errors described in #298, you are introducing many additional Travis jobs per build.  This problem only requires one additional job.

.travis.yml
```diff
@@ -16,6 +16,7 @@ cache:
 env:
   - USE_OPTIONAL=true
   - USE_OPTIONAL=false
+  - USE_SIX_LOWEST_VERSION=1.9 USE_OPTIONAL=true
```
jayvdb: `USE_SIX_LOWEST_VERSION` is very long.  Could be shorter.

amorde: Changed to `SIX_VERSION`

amorde:
Hey @jayvdb, any updates on this?

jayvdb:
@amorde , I am not a committer here.  I can only do code review, which I have done.
I am also waiting for this to be reviewed/merged so that I can complete https://github.com/html5lib/html5lib-python/pull/294. ;-)

amorde:
@jayvdb gotcha, haha sorry for the mixup!

gsnedders:
Thanks for the reviewing, @jayvdb! :)

Ultimately, we should almost certainly declare (and test) minimum required versions for all our dependencies.

PR #309, opened 2016-10-30
Drop Python 2.6 support

di:
Yesterday marks 3 years since the last release of Python 2.6! 🎉

To celebrate, I'm attempting to drop support for it from 156 prominent Python packages (one for every week it's past end-of-life)--including this one!

I've tried my best to remove as much 2.6-specific cruft as I can, but at the very least, this PR will remove the `'Programming Language :: Python :: 2.6'` trove classifier from this projects `setup.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
--- a/README.rst
+++ b/README.rst
@@ -90,7 +90,7 @@ More documentation is available at https://html5lib.readthedocs.io/.
 Installation
 ------------
 
-html5lib works on CPython 2.6+, CPython 3.3+ and PyPy.  To install it,
+html5lib works on CPython 2.7, CPython 3.3+ and PyPy.  To install it,
 use:
 
 .. code-block:: bash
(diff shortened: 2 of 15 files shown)


gsnedders:
As a package vendored by pip, we're not going to drop support till pip does to avoid pip having to maintain a fork, currently planned for pip 10.0 (https://github.com/pypa/pip/issues/3955).

di:
@gsnedders Thanks, it seems this PR is a bit premature then. Sorry for the noise!

gsnedders:
@di I'm more than happy to keep the PR open, because it saves duplicating the work in the probably relatively near future, and I'm grateful for you doing it! It sounds like pip 9.0 is nearer than I expected, so probably nearer to being able to land this than I expected!

jdufresne:
Should this PR be reconsidered now that pip has officially deprecated 2.6 and will be dropping it in the next version?

<https://mail.python.org/pipermail/distutils-sig/2016-November/029785.html>

gsnedders:
The outcome of the discussion I had with @dstufft before was that:

1. The next version will support 2.6 (it is almost certain to be 9.1)
2. 9.x is likely to be the current major release until mid-2017
3. It'd be ideal to keep supporting 2.6 until pip's support is finally dropped (e.g., in case of any last-minute security releases shortly before 10.0)

Given the relatively small gains from dropping 2.6 (as I think the PR shows, the majority of the gains are in being able to finally use set and dict literals), I'm in no rush and happy to wait for 10.0 to ship.

Issue #282, opened 2016-07-18
fix version numbering


phillyc:
I just spent 4 hours with another engineer trying to figure out why our builds weren't producing the same results.

Turns out I had html5lib "0.9999999" and he had "0.999999999" which was nearly impossible to spot on the pip list readout. >.<

http://semver.org/#why-use-semantic-versioning

I'm going to guess that you'll say html5lib isn't ready for production so your versioning scheme is appropriate, but I'm still filing this bug because I'm steamed about losing so much development time. :/

shaileshahuja:
I didn't even realize I was facing this problem until I read this issue. I was wondering how are the same versions behaving differently!

phillyc:
I should point out that they don't appear to to be the same versions. I'm running "0.9999999" which is seven nines. While my coworker is running "0.999999999" which is *nine* nines. :/

gsnedders:
Alas, it's a bit hard to change now.

For those wondering how we ended up here, in September 2013 when we moved to asymptotically approaching 1.0 I expected a release of 1.0 by the end of 2013. Alas, that didn't happen, for a variety of reasons (primarily the spec having some large changes and not wanting to ship something out of date, and lack of time for me and lack of other contributors). It's worthwhile pointing out that html5lib in its origins was essentially a research project and as a support tool to those suggesting changes to the spec, as such it had some weird features and some weird oddities in it.

Now, perhaps 0.99999999 should have been 0.100000000 to clarify the changes, but that's also a confusing version number.

Also in hindsight, we probably should've shipped one of the releases in 2013 as 1.0.

Ultimately, what's blocking shipping 1.0 now is having something up-to-date with the spec and having documentation.

phillyc:
Maybe a simple solution would be to use 0.9.7 -> 0.9.9 -> 0.9.10 ?? 

Then whenever you feel comfortable flipping to 1.0.0 it's a simple and easy read?

gsnedders:
> Maybe a simple solution would be to use 0.9.7 -> 0.9.9 -> 0.9.10 ??

But 0.9.7 is lower version than 0.99 (yet alone the current version!), hence that doesn't work.

ThibWeb:
If this project aims to follow SemVer (#278), the minor version number is supposed to be incremented by 1 for each minor release. What should have happened is 0.9, 0.10, 0.11.

[...]

PR #311, opened 2016-12-02
Bump version to something readable + semver

jpic:
Counting the 9s is a bit tiring, particularely when debugging version conflicts. I suppose this is because html5lib is not judged mature enough to be released as 1.0. This patch proposes moving from 0.999999 to 0.1.0 and at the same time proposes to adopt the semver standard.


gsnedders:
0.1.0 < 0.9999999999, so that doesn't work, so I'm just closing this; discussion should probably happen in #282. The only way out of this wormhole at this point is 0.10000000000 or similar (I didn't count the zeros!) and then ignore all but the trailing zero (i.e., 0.10000000001 as the following release); as mentioned in the other issue, the right approach is almost certainly to just actually document what the public API is and release a 1.0 soon thereafter.

PR #318, opened 2017-02-18
make compatible with python 3.6

moben:
errors on invalid escape sequences in strings, such as "\s"

i.e. anything not mentioned in https://docs.python.org/3/reference/lexical_analysis.html#strings


gsnedders:
Oh, this isn't compatibility per-se, this is just avoiding DeprecationWarnings? (So it's essentially compatibility with some future version.)

moben:
yes. Looking again I might have run into it with `-Werror` I guess.

Still a good idea, though for a future python version.

gsnedders:
Oh, totally, still a good idea; thanks for the patch!

gsnedders:
(In case it was unclear: I didn't merge it earlier because I was still waiting on Travis CI to finish running against the merge with the Travis fix.)

PR #319, opened 2017-02-18
Fix Travis

gsnedders:


gsnedders:
okay, we appear to be hitting some different pip/setuptools bug compared with before, le sigh

gsnedders:
Okay, so we're running into https://github.com/pypa/pip/pull/2616 which was fixed in 6.1.0.

PR #320, opened 2017-02-19
Give minimum versions of all dependencies and test on them

gsnedders:


gsnedders:
This doesn't work because `PIP_VERSION` is consumed by PIP. Need to rename the env variable.

willkg:
I understand the gist of what this is trying to solve, but I think this is better solved with tox and environment matrixes.

I think we should close this out and reimplement it after the PR switching to tox has landed.

Regardless, I'm taking this out of the 1.0 milestone.

ambv:
It would be nice to test the minimum versions of dependencies that we support. Luckily, there really only is `six` and `webencodings` that are required. `six` will go away when Python 2 support is gone.

However, we should NOT be pinning *maximum* versions as overly restrictive version caps cannot be fixed by the user. See: https://iscinumpy.dev/post/bound-version-constraints/

In any case, this PR is quite outdated. I'll make a new one to declare minimum versions of all dependencies that we still use in `master`.

PR #323, opened 2017-02-23
Fix tox.ini to allow specifying pytest tests

willkg:
This fixes tox.ini so that you can pass arguments to py.test.

For example:

```
    tox -e py34-base html5lib/tests/test_serializer.py
```

will now only run the serializer tests.

```
    tox -e py34-base -- -vv
```

will run with super verbose mode and tons of data will go flying by in your
terminal.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,5 +13,5 @@ deps =
   optional: -r{toxinidir}/requirements-optional.txt
 
 commands =
-  {envbindir}/py.test
+  {envbindir}/py.test {posargs}
   {toxinidir}/flake8-run.sh


willkg:
This doesn't affect travis, but it does make development iterations much faster.

Another fix we can do in the tox file is to move the flake8 pass to a separate "environment" so that it runs once in a tox run and not once for every environment.

Does that sound helpful? If so, I can make that change and tack it on to this one.

willkg:
This is super minor. It's just an ergonomic issue related to development.

Issue #183, opened 2015-04-14
Missing MathML elements and attributes


fred-wang:
html5lib/sanitizer.py misses some MathML elements and attributes. In particular the semantics/annotation/annotation-xml elements required to fix a very old planet bug (http://www.maths-informatique-jeux.com/blog/frederic/?post/2013/01/22/Analysis-of-Lithium-s-algorithm#c142).

See https://github.com/html5lib/html5lib-python/pull/181

PR #331, opened 2017-04-10
Add missing MathML tags for annotations. #183

fred-wang:
New attempt, rebasing on top of master and restricting to annotations.

https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics
https://www.w3.org/TR/MathML/chapter3.html#presm.semantics


willkg:
The sanitizer is built to make sure the output is safe regardless of what the input was. So in order to approve adding these, I need to feel confident that these are safe to add.

There's this comment which you wrote that talks about the safety of adding these bits:

https://github.com/html5lib/html5lib-python/pull/181#issuecomment-137714933

However, I don't know anything about MathML and can't evaluate that, so I can't approve this. We'll need someone who understands this to yay/nay it preferably with some proof. I'll hunt around and see if I can find someone I know.

willkg:
I can't find anyone. I'm going to nix this from the 1.0 milestone.

fred-wang:
@willkg What's the plan on this? Did you ask @jruderman ?

willkg:
@fred-wang I didn't find anyone, so I nixed it from the 1.0 milestone. After I got 1.0 out, I stepped off of this project.

PR #332, opened 2017-04-15
Update docs

twm:
When trying to figure out how to sanitize some HTML I noticed that the docs have lagged behind the implementation. So I removed the mention of `HTMLSanitizer` and then... things got out of hand. Summary of changes:

* Add `tox -e doc` environment so that it's easy to build the docs.
* Remove documentation of HTMLTokenizer, as it is now private.
* Remove documentation of HTMLSanitizer, as it no longer exists.
* Add basic documentation of the treeadapters package.
* Linkify many references to classes, functions, and modules.
* Fix various Sphinx warnings and formatting issues.
* Add `__version__` to `html5lib.__all__`
* Remove sub-modules from `html5lib.treeadapters.__all__` (as this caused Sphinx warnings and didn't really make sense).

I didn't squash as suggested in CONTRIBUTING.md because recent PRs don't seem to be following that procedure. I am happy to do so if you like, or to try to break this into smaller PRs.


willkg:
The sanitizer still exists, but it got rewritten as a filter and is now in `html5lib.filters.sanitizer`. You can see it here:

https://github.com/html5lib/html5lib-python/blob/17499b9763a090f7715af49555d21fe4b558958b/html5lib/filters/sanitizer.py

Just a drive-by comment on the off chance it helps.

twm:
@willkg Yup, and the new form is documented. The old docs just weren't removed.

html5lib/__init__.py
```diff
@@ -19,7 +28,8 @@
 from .serializer import serialize
 
 __all__ = ["HTMLParser", "parse", "parseFragment", "getTreeBuilder",
-           "getTreeWalker", "serialize"]
+           "getTreeWalker", "serialize", "__version__"]
```
willkg: I don't think we want `__version__` to get imported when importing `*`.

html5lib/__init__.py
```diff
@@ -19,7 +28,8 @@
...
 
 # this has to be at the top level, see how setup.py parses this
+#: Distribution version number, which asymptotically approaches 1.
```
willkg: I'd remove this. Amongst other things, it's not going to be true for much longer.

doc/movingparts.rst
```diff
@@ -110,11 +105,11 @@ You can alter the stream content with filters provided by html5lib:
   the document
 
 * :class:`lint.Filter <html5lib.filters.lint.Filter>` raises
-  ``LintError`` exceptions on invalid tag and attribute names, invalid
+  :exc:`AssertionError` exceptions on invalid tag and attribute names, invalid
```
willkg: Is it really an AssertionError? If so, we should write up an issue to change that.

doc/movingparts.rst
```diff
@@ -90,15 +86,14 @@ The serializer lets you write HTML back as a stream of bytes.
...
-documentation.
+You can customize the serializer behaviour in a variety of ways. Consult
+the :class:`~html5lib.serializer.HTMLSerializer` documentation.
 
 
 Filters
 ~~~~~~~
 
-You can alter the stream content with filters provided by html5lib:
+html5lib provides several filters
```
willkg: Given that what follows is a bulleted list, can you add a `:` to the end of this line?

doc/movingparts.rst
```diff
@@ -55,17 +52,16 @@ The implementation of builders can be found in `html5lib/treebuilders/
...
 
 The implementation of walkers can be found in `html5lib/treewalkers/
 <https://github.com/html5lib/html5lib-python/tree/master/html5lib/treewalkers>`_.
 
-Walkers make consuming HTML easier. html5lib uses them to provide you
-with has a couple of handy tools.
+html5lib provides a few tools for consuming token streams:
 
+* :class:`~html5lib.serializer.HTMLSerializer`, to generate a stream of bytes; and
+* filters, to manipulate the token stream.
```
willkg: The leader has "a few tools", but this list has two items. Further, the items seem like a sentence. I'd either change the bullet list into a sentence or unsentencify the bullet items. Maybe something like this?:

```
html5lib provides two tools for consuming token streams:

* :class:`~html5lib.serializer.HTMLSerializer` for generating a stream of bytes
* filters for manipulating the token stream
```

doc/html5lib.rst
```diff
@@ -1,13 +1,8 @@
 html5lib Package
 ================
 
-:mod:`html5lib` Package
------------------------
```
willkg: Why take the header out here?

html5lib/__init__.py
twm: As `__version__` is part of the public interface of this module, which `__all__` defines, I think it should be here.

doc/html5lib.rst
twm: There are two headers in a row with exactly the same text.

doc/movingparts.rst
twm: I ended up going with a sentence, rather than the bulleted list, as there aren't exactly two (really there's the serializer and a bunch of filters), so it's really two _categories_ of token consumer, but that distinction isn't really useful to point out.

html5lib/__init__.py
twm: Thank goodness! Will do.

doc/movingparts.rst
twm: Yeah, the implementation is basically all `assert` statements: https://github.com/html5lib/html5lib-python/blob/7bbde541e08ca82c07af5ac7e5f64525f1793814/html5lib/filters/lint.py#L24-L79

twm:
I think that I have addressed all the issues you noted as appropriate. Please let me know if anything else is required! Thanks!

html5lib/__init__.py
willkg: I wouldn't think about `__all__` that way. It's really for specifying what gets exported if the user does `from html5lib import *`. I don't want `__version__` to get exported in that scenario. Please undo this change.

willkg:
@twm Thank you so much for this!

PR #364, opened 2017-11-23
Re-enable codecov

jayvdb:
This also removes dependence on `./requirements-install.sh` and `flake8-run.sh`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,21 +1,23 @@
 [tox]
-envlist = {py27,py33,py34,py35,py36,pypy}-{base,optional}
+envlist = {py27,py33,py34,py35,py36,pypy}-{base,six19,optional}
 
 [testenv]
 deps =
-  flake8<3.0
-  pytest
-  pytest-expect>=1.1,<2.0
-  mock
-  base: six
-  base: webencodings
   optional: -r{toxinidir}/requirements-optional.txt
+  -r{toxinidir}/requirements-test.txt
   doc: Sphinx
 
+passenv =
+  PYTEST_COMMAND
+  COVERAGE_RUN_OPTIONS
 commands =
-  {envbindir}/py.test {posargs}
+  six19: pip install six==1.9
+  {env:PYTEST_COMMAND:{envbindir}/py.test} {posargs}
   flake8 {toxinidir}
 
 [testenv:doc]
 changedir = doc
 commands = sphinx-build -b html . _build
+
+[flake8]
+exclude = ./.tox
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,6 +2,7 @@
 environment:
   global:
     PATH: "C:\\Python27\\Scripts\\;%PATH%"
+    PYTEST_COMMAND: "coverage run -m pytest"
   matrix:
     - TOXENV: py27-base
     - TOXENV: py27-optional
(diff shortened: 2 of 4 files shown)


jayvdb:
(Once CI finishes; I'll rebase to pull in the other CI patch)

jayvdb:
https://github.com/html5lib/html5lib-python/pull/365 should be merged first, as that makes this one a bit easier.

willkg:
I'll take a look at this tomorrow. Thank you!

tox.ini
```diff
@@ -1,21 +1,23 @@
...
 deps =
-  flake8<3.0
-  pytest
-  pytest-expect>=1.1,<2.0
-  mock
-  base: six
-  base: webencodings
   optional: -r{toxinidir}/requirements-optional.txt
+  -r{toxinidir}/requirements-test.txt
   doc: Sphinx
```
willkg: It looks like when this ran the tests with `six==1.9`, it'd also run the tests with the optional requirements. I'm pretty sure your changes don't do that--`six19` is a separate environment from `optional`.

I don't know offhand if that's *important*, but it seems like if we want to maintain parity, it'd be better to add a line here like:

```
six19: -r{toxinidir}/requirements-optional.txt
```

tox.ini
```diff
@@ -1,21 +1,23 @@
...
 
+passenv =
+  PYTEST_COMMAND
+  COVERAGE_RUN_OPTIONS
 commands =
-  {envbindir}/py.test {posargs}
+  six19: pip install six==1.9
+  {env:PYTEST_COMMAND:{envbindir}/py.test} {posargs}
   flake8 {toxinidir}
```
willkg: In other projects I work on, I run flake8 as a separate environment because there isn't much reason to run it multiple times. Having said that, it seems like it only takes a few seconds on html5lib-python, so seems like it'd not a big deal here.

willkg:
@jayvdb Are you still working on this or are you done?

tox.ini
jayvdb: flake8 could be moved out to a separate factor easily, and even a different testenv.

It is important to run flake8 on each version of python, as pyflakes is version dependent.

The benefit of keeping it in here is that devs dont need to remember to do it, and they typically only test on one environment.  For CI, the running time is inconsequential compared to the rest of the very voluminous tests ;-)

tox.ini
jayvdb: Sorry, yes I missed that.  It would be better to achieve that by using `TOXENV=six19-optional` .

jayvdb:
You've found one thing that needs fixing, and I'm happy to change the flake8 invocation also.

tox.ini
willkg: +1 to everything you say here. Don't change what you have.

willkg:
@jayvdb Can you pin pytest to version 2.2.5? I think that'll make this PR work better in CI and I'll just drop PR #372.

jayvdb:
2.2.5 , or 3.2.5?

willkg:
Bleh. I'm having typing problems today. I meant 3.2.5. Thanks!

jayvdb:
I have cherry-picked your commit; I hope that is OK.

willkg:
You rock--thank you!

PR #365, opened 2017-11-23
Split AppVeyor build into 10 jobs

jayvdb:
The AppVeyor build running all tox environments in
a single job takes a long time to report any status,
and the log is difficult to load.


jayvdb:
It doesnt appear to be any slower; could even be quicker.

willkg:
I'll take a look at this tomorrow. Thank you!

Issue #370, opened 2017-11-28
Drop six 1.9 support


gsnedders:
The version of `pytest` on Travis's 3.6 environment seems to rely on `six>=1.10` so our `six==1.9` test config fails. See, e.g., https://travis-ci.org/html5lib/html5lib-python/jobs/308631931

Given 1.10 is now over two years old, I'd hope we can drop support for 1.9, and fix this by just bumping our min required version to 1.10.

gsnedders:
(see #301, #298 for the 1.9 requirement)

willkg:
I wonder if this isn't related to Travis, but rather that we don't specify which version of pytest to install and use:

https://github.com/html5lib/html5lib-python/blob/master/requirements-test.txt#L4

I forget if that file affects what gets installed, but can look into it in a couple of hours or tomorrow.

hugovk:
Whatever the cause, this is no longer happening, the 3.6 env is passing on Travis (and all of them: 2.7, 3.5-3.8, PyPy and PyPy3).

Latest master build:

* https://travis-ci.org/html5lib/html5lib-python/builds/655163448

Can this issue now be closed?

---

In case you do want to bump to six 1.10, here's a passing branch I can PR:

* https://travis-ci.org/hugovk/html5lib-python/builds/655374557

(And https://github.com/html5lib/html5lib-python/pull/429 and https://github.com/html5lib/html5lib-python/pull/441 will bump to latest pytest 5.)

ambv:
Currently html5lib works with six 1.9+, including the latest 1.16.0 (example: https://ci.appveyor.com/project/gsnedders/html5lib-python/builds/46378000/job/kfed1o7y9uw02q9i). We can close this.

PR #372, opened 2017-11-29
Pin pytest to a version that doesn't require six

willkg:
pytest 3.3.0 added `six>=1.10.0` as a requirement which messes us up. Either we should drop support for six 1.9 or we can land this which pins pytest to a slightly earlier version as a stopgap.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,6 +1,6 @@
 -r requirements.txt
 
 flake8<3.0
-pytest
+pytest==3.2.5
 pytest-expect>=1.1,<2.0
 mock


willkg:
Pretty sure this gets around the problem that prompted issue #370. Though we should do that as well at some point.

@gsnedders When Travis and AppVeyor are happy, can you review this?

willkg:
There's another PR that fixes `tox.ini` to use `requirements-test.txt`. I could pull that change into this one, but then that one will require more work to merge. I'm inclined to leave this as is and let that one fix `tox.ini`.

That's the root cause of the AppVeyor failure.

jayvdb:
Fyi, I'll be able to help in about 15 mins.

jayvdb:
I've rebased & fixed my PR (https://github.com/html5lib/html5lib-python/pull/364), and if you rebase this PR on top of mine, I think everything should be green.

willkg:
PR #364 pins pytest now. I'm going to close this PR out.

PR #385, opened 2017-12-04
Speed up CI

hugovk:
It's usually quicker to have the slower build jobs run first, that way the build isn't waiting as at the end for one or two slow ones to cmoplete.

Taking numbers from a master build, rounding to nearest minutes, gives something like this in theory for five parallel jobs:

```
xxx
xx
xxxx
xxx
xxx
  xxx
   xxxx
   xxx
   xxx
    xxx
     xx
      xx
      xx
       xx
       xxx
       xxx
        xxx
        xxx
         xxxxxxx
          xxxxxxxxx
          xxxxxxxxx
1234567890123456789
```

In reality:
* Ran for 26 min 43 sec
https://travis-ci.org/hugovk/html5lib-python/builds/310874897

Slowest first, using same numbers:
```
xxxxxxx
xxxxxxxxx
xxxxxxxxx
xxx
xxx
   xxx
   xx
     xx
      xxx
       xxx
       xx
         xx
         xxxx
         xxx
         xxx
          xxx
           xxx
            xxx
            xxx
             xx
             xxxx
12345678901234567
```

This PR build:
* Ran for 16 min 7 sec
https://travis-ci.org/hugovk/html5lib-python/builds/309593149


willkg:
I looked at the master builds for the last 10 rounds and the numbers are kind of all over the place and not consistent. So I'm not sure I see evidence that this will really improve the situation.

The specific change here is just reordering the environments which is benign, so I'm going to approve it and land it and we can see how future builds affect the hypothesis.

hugovk:
For reference, here's the last 10 passing master builds (685-711), latest first:

Ran for 20 min 45 sec
Ran for 26 min 58 sec
Ran for 22 min 12 sec
Ran for 24 min 32 sec
Ran for 17 min 33 sec
Ran for 19 min 5 sec
Ran for 23 min 14 sec
Ran for 16 min 15 sec
Ran for 17 min 58 sec
Ran for 14 min 58 sec

Issue #203, opened 2015-08-17
Update void elements


annevk:
https://github.com/html5lib/html5lib-python/blob/master/html5lib/constants.py#L487 is no longer accurate. Seems to miss `keygen`, `wbr`, `menuitem`.

And seems to have `event-source` which is no longer a thing.

jayaddison:
@willkg @gsnedders fyi - it looks like https://github.com/html5lib/html5lib-python/pull/395 can contribute to resolving this issue.

NB: The code block referenced in the issue description within `constants.py` is now viewable [here](https://github.com/html5lib/html5lib-python/blob/af19281fa28788830684b4c8fc6d0c588b092616/html5lib/constants.py#L559-L575).

ambv:
Looks like `menuitem` [is deprecated, and moreover not really void](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem).

`keygen` [is also deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen).

So the only missing one is `wbr`.

I don't think it's worth adding `keygen` as it's already deprecated and was never used in html5lib. Moreover, I think we could rather uncontroversially remove `command` ([removed entirely from the spec in 2012](https://github.com/whatwg/html/commit/9e2e25f4ae90969a7c64e0763c98548a35b50af8)), and `event-source` ([renamed to `eventsource` in July 2008](https://github.com/whatwg/html/commit/d157945d0285b4463a04b57318da0c4b300a99e7) and [dropped entirely in February 2009](https://github.com/whatwg/html/commit/43cbdbfbb7eb74b0d65e0f4caab2020c0b2a16ff)).

PR #395, opened 2018-06-10
Support the <wbr> element

twm:
``<wbr>`` is an empty element which indicates a line break opportunity. The effect is similar to inserting a zero-width space, but superior because ``<wbr>`` does not affect the text of the selection (so when copied there aren't invisible characters in it).

``<wbr>`` is safe when sanitizing HTML.

Partially addresses #203.

https://html.spec.whatwg.org/#the-wbr-element


twm:
@gsnedders Is there anything blocking merge of this or #396 on my end?

PR #401, opened 2018-09-14
Remove unnecessary compatibility shim for BytesIO

jdufresne:
`io.BytesIO` is available on all supported Pythons. Can simplify by assuming the import will succeed.


jdufresne:
The AppVeyor failures are unrelated to this changed and are fixed in #397.

PR #404, opened 2018-10-11
Replace deprecated optparse with argparse

jdufresne:
https://docs.python.org/3/library/optparse.html

> Deprecated since version 3.2: The optparse module is deprecated and
> will not be developed further; development will continue with the
> argparse module.

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python
-"""usage: %prog [options] filename
-
+"""
 Parse a document to a tree, with optional profiling
 """
 
+import argparse
 import sys
 import traceback
-from optparse import OptionParser
 
 from html5lib import html5parser
 from html5lib import treebuilders, serializer, treewalkers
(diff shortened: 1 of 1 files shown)


parse.py
```diff
@@ -152,91 +151,83 @@ def printOutput(parser, document, opts):
 
 
 def getOptParser():
```
hugovk: Is it worth renaming this to (for example) `getArgParser`, and have `getOptParser` call the new one with a deprecation warning?

jdufresne: I agree. I've named it `get_parser()` as I don't think the name needs to embed the implementation choice.

I didn't add a compatibility shim as I don't believe this is a public interface. But if I'm wrong and the project maintainers would like to see a shim here, please let me know and I'll ad it.

hugovk:
Merged into the fork here: https://github.com/html5lib-python/html5lib-python/pull/3

Issue #402, opened 2018-09-21
DeprecationWarning in collections import


tirkarthi:
Installing html5lib on CPython master generates the below warning. I think this was added as part of Python 3.7 and will be removed in the next version 3.8 which is under development. Just wanted to give a heads up on this.

```
/home/karthi/requests-html-env/lib/python3.8/site-packages/pip/_vendor/html5lib/_trie/_base.py:3: 
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping
```

Relevant code : https://github.com/html5lib/html5lib-python/blob/950ea0ec60970ba1c88bd9e35d8d85c2a1f318e4/html5lib/_trie/_base.py#L3

Possible solution : https://github.com/pyparsing/pyparsing/pull/16

Thanks

rzuckerm:
When is this change going to be released? I'm still seeing this warning.

PR #408, opened 2018-12-22
Fix collections imports for Python 3.8

### LINKED ISSUE #405: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated
It will stop working in Python 3.8.

labrys:
Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated

Fixes #405

### DIFF
--- a/html5lib/_tokenizer.py
+++ b/html5lib/_tokenizer.py
@@ -2,8 +2,6 @@ from __future__ import absolute_import, division, unicode_literals
 
 from six import unichr as chr
 
-from collections import deque
-
 from .constants import spaceCharacters
 from .constants import entities
 from .constants import asciiLetters, asciiUpper2Lower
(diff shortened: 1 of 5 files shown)


methane:
Could you make an release?
pip bundles html5lib which version is released on PyPI.

olf42:
Hey @gsnedders, as Python 3.8 is released, I guess it would be generally very much appreciated if you could make a releasing a version which includes this fix. Thank you very much!

eli-schwartz:
There is no fix here, this pull request makes no sense and was retracted by its author in response to this comment: https://github.com/html5lib/html5lib-python/issues/405#issuecomment-449598616

The actually relevant discussion was at #402 and #403, and #419 is already submitted asking for a new release.

olf42:
I already noticed, but thanks for pointing it out. I forgot to delete my useless comment. :/

PR #412, opened 2019-02-21
Allow data: to not be part of sanitizer’s allowed protocols

charmander:
When `data:` wasn’t permitted, it could attempt to `del attrs[attr]` twice.


twm:
The change looks valid, but shouldn't there be a test case for this?

PR #413, opened 2019-03-07
max-width and min-width added to allowed_css_properties

kuvandjiev:
max-width and min-width added to allowed_css_properties to improve style sanitization.


ambv:
Superseded by #566.

Issue #411, opened 2019-02-14
Support pytest 4


FRidh:
I suppose the test suite is not pytest 4.2 compatible.

```py
============================= test session starts ==============================
platform linux -- Python 3.7.2, pytest-4.2.1, py-1.7.0, pluggy-0.8.1
rootdir: /build/html5lib-1.0.1, inifile: pytest.ini
plugins: expect-1.1.0
collected 31264 items / 1 errors / 31263 selected

==================================== ERRORS ====================================
________________ ERROR collecting html5lib/tests/test_stream.py ________________
/nix/store/mljhw8f6ng9ny9ahx68m9fj6bq0ghbqv-python3.7-pluggy-0.8.1/lib/python3.7/site-packages/pluggy/hooks.py:284: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
/nix/store/mljhw8f6ng9ny9ahx68m9fj6bq0ghbqv-python3.7-pluggy-0.8.1/lib/python3.7/site-packages/pluggy/manager.py:68: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
/nix/store/mljhw8f6ng9ny9ahx68m9fj6bq0ghbqv-python3.7-pluggy-0.8.1/lib/python3.7/site-packages/pluggy/manager.py:62: in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
/nix/store/pf8c7km0691nci67allvq2ggsgvr8asl-python3.7-pytest-4.2.1/lib/python3.7/site-packages/_pytest/python.py:226: in pytest_pycollect_makeitem
    res = list(collector._genfunctions(name, obj))
/nix/store/pf8c7km0691nci67allvq2ggsgvr8asl-python3.7-pytest-4.2.1/lib/python3.7/site-packages/_pytest/python.py:403: in _genfunctions
    self.ihook.pytest_generate_tests(metafunc=metafunc)
/nix/store/mljhw8f6ng9ny9ahx68m9fj6bq0ghbqv-python3.7-pluggy-0.8.1/lib/python3.7/site-packages/pluggy/hooks.py:284: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
/nix/store/mljhw8f6ng9ny9ahx68m9fj6bq0ghbqv-python3.7-pluggy-0.8.1/lib/python3.7/site-packages/pluggy/manager.py:68: in _hookexec
[...]

PR #414, opened 2019-03-28
Support pytest 4

hroncok:
Fixes https://github.com/html5lib/html5lib-python/issues/411

This is not optimal. The yield tests should probably be transferred to parametric tests or subtests (needs pytest 4.4).


hugovk:
Please see PR #429 for a rebase of this PR.

PR #417, opened 2019-04-18
Travis CI: Add Python 3.7 and pypy3

cclauss:
Python 3.4 is End of Life: https://devguide.python.org/devcycle/#end-of-life-branches

[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo: false__ in your __.travis.yml__, we recommend removing that configuration_"

Conditionally install __datrie__ only if on CPython < 3.7 as a workaround to pytries/datrie#52 which although ___closed___ is not yet released to https://pypi.org/project/datrie as discussed in pytries/datrie#68.  It would also be possible to add `pip3 install git+https://github.com/pytries/datrie.git` on CPython >= 3.7.


gsnedders:
FWIW: given we're a vendored dependency of pip, we've typically avoided dropping support for any version of Python still supported by pip, which implies we should still support Py3.4.

gsnedders:
Thanks!

PR #420, opened 2019-05-30
Fix Pytest4.x compatibility error

stanislavlevin:
According to pytest docs:
```
marks in pytest.mark.parametrize
Removed in version 4.0.

Applying marks to values of a pytest.mark.parametrize call is now deprecated.
...

This was considered hard to read and understand, and also its implementation
presented problems to the code preventing further internal improvements in the
marks architecture.

To update the code, use pytest.param
```


marv:
What's the status of this? Would be great to have a release that's compatible with `pytest >= 4`

hugovk:
Merged into the fork here: https://github.com/html5lib-python/html5lib-python/pull/7

gsnedders:
@hugovk do you have any views on this v #429?

gsnedders:
Oh, wait, #429 is a superset of this.

PR #421, opened 2019-06-05
Drop support for EOL Python 3.4

hugovk:
Python 3.4 is EOL since 2019-03-16 and no longer receiving security updates (or any updates) from the core Python team.

This allows 3 jobs to be removed from the build, and allows the project to concentrate on supported Python versions.

This PR also adds `python_requires` to help pip install the correct package version for their running Python.


gsnedders:
https://github.com/pypa/pip/commit/2c1010ee96d73daf950ffff66bbd9b6718f2f0e9 dropped support for 3.4 in pip, shipped in 19.2.

PR #292, opened 2016-07-26
Enable more linting

jayvdb:
WIP-ish PR.  Interested to see if there are any parts that are objectionable.

I'm happy to split it into smaller chunks for different types of lint issues, with nice commit messages.

Also, I can add more fine tuning by implementing my own [flake8-putty](https://pypi.python.org/pypi/flake8-putty), which allows rules to be disabled per-file and even per-line without modifying the source, so new rules can be enforced for new code while existing lint errors in old code are ignored.

---
This change is [Reviewable](https://reviewable.io/reviews/html5lib/html5lib-python/292)

PR #428, opened 2019-11-01
Update flake8 to the latest version and fix all errors

jdufresne:


hugovk:
Merged into the fork here: https://github.com/html5lib-python/html5lib-python/pull/1

gsnedders:
FWIW, I have a preference in favour of pinning flake8 and its dependencies to stop tests from starting failing without changes in html5lib, though preferably in combination with adding some bot to update pinned dependencies.

jdufresne:
No problem. I've changed the PR to keep flake8 pinned, but now to the latest version. All errors are still fixed.

hugovk:
> preferably in combination with adding some bot to update pinned dependencies.

The two bots I know about: dependabot, now part of GitHub; and pyupbot, which I hear has been a bit flaky of late.

gsnedders:
Yeah, we use pyupbot on web-platform-tests and it seems to have been going downhill…

html5lib/_utils.py
```diff
@@ -27,7 +27,7 @@
         # We need this with u"" because of http://bugs.jython.org/issue2039
         _x = eval('u"\\uD800"')  # pylint:disable=eval-used
         assert isinstance(_x, text_type)
-except:  # pylint:disable=bare-except
+except Exception:  # pylint:disable=bare-except
```
gsnedders: Comment is now wrong (though maybe if we  #292 not worth keeping around? but I guess we will?).

jdufresne: Thanks for catching. Fixed in the latest revision.

PR #430, opened 2019-11-08
Deploy to PyPI from Travis CI

hugovk:
This should help make releasing easier.

* Deploy non-tagged builds to Test PyPI
* Deploy tagged builds to live PyPI

Will only trigger when run from the `html5lib/html5lib-python` repo, so not for forks.

TODO:

1. Add an API token called (say) `html5lib-ci` at https://test.pypi.org/manage/account/token/
2. On the command line, install the [Travis Client](https://github.com/travis-ci/travis.rb#installation)
3. Run `travis encrypt`
4. Paste in the token from step 1, it begins `pypi-`
5. Ctrl-D to end
6. Add the encrypted value as the `password: secure: "TODO"` value on line 53
7. Repeat for production PyPI, put the encrypted value on line 64

More info on PyPI API tokens: https://pypi.org/help/#apitoken


gsnedders:
> Deploy non-tagged builds to Test PyPI

Perhaps stupid question, "Test PyPI"?

hugovk:
It's "a separate instance of the Python Package Index that allows you to try distribution tools and processes without affecting the real index".

https://test.pypi.org/

I find it useful to deploy to TestPyPI on untagged merges to master. That way, it's actually testing the release process and machinery, to give confidence that tagged builds deploy to production PyPI. 

For example, I have it set up in https://github.com/hugovk/pypistats, which also uses setuptools_scm, so the dev version number automatically updates, and so a new upload can go to TestPyPI for each merge. 

Another benefit is it creates WIP builds which people could use for testing in between releases. 

More info on TestPyPI:
https://packaging.python.org/guides/using-testpypi/

hugovk:
Closing old PR, but let me know if you'd like it re-opened. I'd also use GitHub Actions instead today, as the new Travis CI pricing model isn't making things easy for OSS.

PR #432, opened 2020-01-09
Remove support for end-of-life Python 2.7

jdufresne:
Python 2.7 has been EOL since 2020-01-01. It is no longer receiving
security updates.

By removing Python 2 support, can reduce testing resources and simplify
the code by removing compatibility shims.

Removes dependency on six.

Removes test dependency on mock.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,8 +4,6 @@ environment:
     PATH: "C:\\Python27\\Scripts\\;%PATH%"
     PYTEST_COMMAND: "coverage run -m pytest"
   matrix:
-    - TOXENV: py27-base
-    - TOXENV: py27-optional
     - TOXENV: py35-base
     - TOXENV: py35-optional
     - TOXENV: py36-base
@@ -21,7 +19,7 @@ test_script:
   - tox
 
 after_test:
-  - python debug-info.py
+  - C:\\Python35\\python.exe debug-info.py
 
 on_success:
   - codecov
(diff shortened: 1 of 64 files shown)


jgraham:
I am strongly opposed to this. There are still many consumers of html5lib who can't or haven't migrated to Python 3. Although PSF is no longer supporting Python 2.7 it's still under support in e.g. RHEL.

jdufresne:
No problem. With this change, I'm only trying to help this project and the larger Python community. I figure sooner or later, someone would need to do this maintenance work. If the project policy is to continue supporting Python 2.7, no worries.

That said, Python 2.7 user can continue using the last release of html5lib that supported 2.7 which is 1.0.1 as of this writing, released Dec 7, 2017. As users haven't had a new release in over a year, there wouldn't really be any immediate change for them. The setup.py file users `python_requires` so pip would choose the best compatible version automatically.

gsnedders:
Note that this implies pip staying on the last Py2-supporting release, which is potentially troublesome given the current last Py2 release doesn't support Py3.future and implies pip vending multiple versions of html5lib. I'd rather stick to the somewhat unwritten policy of supporting at least the same versions that pip supports (and we should document that).

Let's move any discussion about the policy to #439, and close this for now.

jdufresne:
Now that [pip has dropped support for Python 2](https://github.com/pypa/pip/commit/be9637f7f57d1d6e97dfbdd2bdc698edbba73d9f), perhaps this PR can be reconsidered? If so, I'll happily rebase.

PR #438, opened 2020-02-24
Test on Python 3.9-dev to avoid surprises

hugovk:
Python 3.9 is in alpha, with beta in May and full release in October.

* https://www.python.org/dev/peps/pep-0596/


hugovk:
Flake8 failures will be fixed by https://github.com/html5lib/html5lib-python/pull/428.

hugovk:
Merged into the fork here: https://github.com/html5lib-python/html5lib-python/pull/8

FelixSchwarz:
just wanted to add that testing on the latest version would be also good for Fedora. We are updating to new Python releases pretty quickly and CPython changed to an annual release cycle which fits better for Linux distros like Fedora.

Most packagers will report errors found during the mass rebuilds when transitioning to new CPython releases but of course it would be much more efficient if these were spotted upstream. :-)

Issue #439, opened 2020-02-25
Document Python version support policy


gsnedders:
It's been an intended goal to match a (non-strict) superset of the versions that [pip supports](https://pip.pypa.io/en/stable/installing/#python-and-os-compatibility). We should probably document this somewhere.

PR #447, opened 2020-02-26
Document Python support

hugovk:
Suggestion to fix https://github.com/html5lib/html5lib-python/issues/439.

Also update list of supported versions, EOL 3.4 was dropped in https://github.com/html5lib/html5lib-python/pull/421 (and pypa/pip@2c1010e dropped support for 3.4 in pip, shipped in 19.2).


hugovk:
Well, it's a start :)

PR #449, opened 2020-02-26
Pin all the test requirements to try and make testing deterministic

gsnedders:


jgraham:
If you're going to do this should you also provide and require hashes?

gsnedders:
> If you're going to do this should you also provide and require hashes?

perhaps, but we've never bothered with doing that for wpt and I'm unaware of it actually causing issues? it bloats the file a lot and makes manual updates harder, no?

jgraham:
I'm not proposing to block on that, but at least at Mozilla it's considered standard practice to require hashes for Python deps to reduce the chance of issues if there's a compromise in PyPI or similar, and typically a bot can do the work of generating a diff for the update.

gsnedders:
@jgraham this breaks with `-r requirements.txt`, unless we pin the requirements we publicly require which… doesn't seem great.

PR #455, opened 2020-02-29
Drop datrie support

gsnedders:
see #442


gsnedders:
Landed given no review forthcoming in three months.

PR #482, opened 2020-04-13
Make scanning for meta encoding much quicker

gsnedders:
Previously, this code tried to match everything with strings
beginning with "<"; now we jump forward to each "<" and compare
there. This also alters the jumpTo implementation to avoid
computing a (perhaps long) slice, making repeated calls O(n^2).

### DIFF
--- a/html5lib/_inputstream.py
+++ b/html5lib/_inputstream.py
@@ -668,15 +668,11 @@ class EncodingBytes(bytes):
     def jumpTo(self, bytes):
         """Look for the next sequence of bytes matching a given sequence. If
         a match is found advance the position to the last byte of the match"""
-        newPosition = self[self.position:].find(bytes)
-        if newPosition > -1:
-            # XXX: This is ugly, but I can't see a nicer way to fix this.
-            if self._position == -1:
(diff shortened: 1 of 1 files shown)


html5lib/_inputstream.py
```diff
@@ -668,15 +668,11 @@ def matchBytes(self, bytes):
...
-        newPosition = self[self.position:].find(bytes)
-        if newPosition > -1:
-            # XXX: This is ugly, but I can't see a nicer way to fix this.
-            if self._position == -1:
-                self._position = 0
-            self._position += (newPosition + len(bytes) - 1)
-            return True
-        else:
+        try:
+            self._position = self.index(bytes, self.position) + len(bytes) - 1
```
jgraham, line 672: Did you test the tradeoff between `.find()` here and `.index()`? The former could avoid constructing an exception in the no-match case, but will need some additional logic.

gsnedders: `index` is a minor win; it's pretty unusual there's no match, and it avoids the branch in the interpreted code in every case.

PR #502, opened 2020-06-10
Don't use cElementTree on Python 3

### LINKED ISSUE #494: Avoid cElementTree on Python 3
`cElementTree` has been [deprecated since 3.3](https://docs.python.org/3/library/xml.etree.elementtree.html), and is simply an alias to `ElementTree` right now. It [will be removed in the Python 3.10](https://bugs.python.org/issue36543).

Originally filed as: pypa/pip#8278
See also: python/cpython#19921

gsnedders:
This just directly imports the first commit from https://github.com/pypa/pip/pull/8278. Fixes #494. This doesn't fill me with joy, but I guess sometimes we do the perhaps silly thing for pip.

(@tiran this is just your commit, let me know if you have any objection to it landing here)

### DIFF
--- a/html5lib/_utils.py
+++ b/html5lib/_utils.py
@@ -7,12 +7,15 @@ try:
 except ImportError:
     from collections import Mapping
 
-from six import text_type
+from six import text_type, PY3
 
-try:
-    import xml.etree.cElementTree as default_etree
-except ImportError:
+if PY3:
     import xml.etree.ElementTree as default_etree
+else:
+    try:
+        import xml.etree.cElementTree as default_etree
+    except ImportError:
+        import xml.etree.ElementTree as default_etree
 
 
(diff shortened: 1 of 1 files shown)


gsnedders:
> r+ but I agree with the preference for `if not PY2` in this case, although I'm obviously hoping Python 4 isn't a real thing since it will just cause more needless breakage.

Note that this has been discussed upstream at six on a number of occasions: see https://bitbucket.org/gutworth/six/issues/22/variable-sixpy3-is-time-bomb and https://github.com/benjaminp/six/pull/297, as well as the python-dev thread at https://mail.python.org/archives/list/python-dev@python.org/thread/M5N3PZP4IUMKRZJKKJQA4IXK5VDX5MOP/ about 3.10 v. 4.

gsnedders:
I'm landing this as-is because if that's what a CPython core developer is doing that's good enough for me 🤷

PR #510, opened 2020-09-30
Improving it to make it more readable requirements-test.txt to requirements_test.txt

Chirag0410:


hugovk:
How is this more readable? This also breaks the build. Is this hacktoberfest spam?

gsnedders:
It's definitely spam, whatever it is.

PR #516, opened 2020-10-27
Import collection ABCs from collections.abc

monoidic:
`DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 
3.3, and in 3.9 it will stop working`

### DIFF
--- a/html5lib/_tokenizer.py
+++ b/html5lib/_tokenizer.py
@@ -2,7 +2,11 @@ from __future__ import absolute_import, division, unicode_literals
 
 from six import unichr as chr
 
-from collections import deque, OrderedDict
+from collections import deque
+try:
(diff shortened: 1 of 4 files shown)


gsnedders:
Um, `OrderedDict` isn't in `collections.abc` in any version of Python? Look at https://github.com/python/cpython/blob/3.9/Lib/_collections_abc.py for the Python 3.9 version of `collections.abc`?

monoidic:
Uhh. Whoops. I definitely wasn't paying enough attention to what I was doing, apparently I was using version 1.0.1 of html5lib when I got that warning and it was for `Mapping`, not `OrderedDict`, but that's been fixed already.
Closing.

PR #519, opened 2020-12-29
Consistency: consume a single character at a time during attribute name state

jayaddison:
This is a small consistency fixup relating to the way that attribute names are retrieved; it also makes some follow-up refactoring work a little cleaner.

Parsing continues fine if we consume a single character at a time during attribute name tokenization, and this doesn't appear to affect performance positively or negatively.


gsnedders:
> it also makes some follow-up refactoring work a little cleaner

FWIW, I think it's pretty likely that any Cython-compiled version of html5lib, once that exists, will use `charsUntil` more widely than we do today.

jayaddison:
> > it also makes some follow-up refactoring work a little cleaner
> 
> FWIW, I think it's pretty likely that any Cython-compiled version of html5lib, once that exists, will use `charsUntil` more widely than we do today.

That's a good goal/consideration to keep in mind, thanks.  For this instance, the suggested change is largely to help indicate that there's no accidental change-of-behaviour introduced by the refactoring in https://github.com/html5lib/html5lib-python/pull/521/files#diff-84be0df9e74521d407f26e2277a2c70be21dbe6012fea9a5786721c5027e2cfaL894-R868 

It also seems consistent with the [comment and logic in `tagNameState`](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/html5lib/_tokenizer.py#L451-L453) (I didn't copy the comment over - but could do)

jayaddison:

PR #521, opened 2020-12-30
Tokenizer: use Python objects to represent tokens

### LINKED ISSUE #24: Move tokens to being objects

jayaddison:
This change refactors the tokenizer module to use Python object instances where previously plain dictionaries were used to hold token state.

This builds upon #519, #520 and attempts to resolve #24.


jayaddison:
NB: This isn't suitable for merge currently; it seems to introduce a noticeable performance penalty:

**Before** (2c19b98)
```
html_parse_etree: Mean +- std dev: 202 ms +- 10 ms
```

**After** (8772408)
```
html_parse_etree: Mean +- std dev: 226 ms +- 11 ms
```

gsnedders:
> This builds upon #519, #520 and attempts to resolve #24.

Leaving reviewing this till after those two, FYI.

jayaddison:

PR #522, opened 2020-12-31
Refactor: unpack and re-order HTMLParser.mainLoop foreign-content conditionals

jayaddison:
This changeset should functionally be a no-op, and may be best reviewed with whitespace changes ignored since the indentation level for some of the loop's logic has been reduced.

The changes appear to result in a minor performance improvement; that said there are likely larger benefits to be found by further simplification and refactoring (perhaps including fairly substantial logical restructuring).

**Before** (2c19b9899ab3a3e8bd0ca35e5d78544334204169)
```
html_parse_etree: Mean +- std dev: 204 ms +- 10 ms
```

**After** (369a4123ae453845a383dafb82e7459a4e87839b)
```
html_parse_etree: Mean +- std dev: 197 ms +- 9 ms
```


html5lib/html5parser.py
```diff
@@ -205,45 +205,43 @@ def mainLoop(self):
...
-                          type in (CharactersToken, SpaceCharactersToken))) or
-                        (currentNodeNamespace == namespaces["mathml"] and
-                         currentNodeName == "annotation-xml" and
-                         type == StartTagToken and
-                         token["name"] == "svg") or
-                        (self.isHTMLIntegrationPoint(currentNode) and
-                         type in (StartTagToken, CharactersToken, SpaceCharactersToken))):
+                    break
+
+                prev_token = new_token
```
jayaddison, line 213: NB: This is technically a behaviour change since `prev_token` could previously have referenced a `ParseErrorToken`.  That said, I don't believe that the [code that inspects the `prev_token`](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/html5lib/html5parser.py#L248-L251) would ever be relevant for error tokens.

jayaddison:

PR #523, opened 2020-12-31
Tree Construction: do not use phase-specific handling for implied end-tag tokens emitted during InBody phase

### LINKED ISSUE #111: li in foster parent case incorrectly handled
The result of parsing by html5lib for the following fragment is not compatible with the specification.

&lt;table>&lt;li>&lt;li>&lt;/li>&lt;/table>

It is parsed into the following DOM tree.

<pre>
&lt;li>
&lt;table>
   &lt;li>
</pre>

It should be parse as the following DOM tree.

<pre>
&lt;li>
&lt;li>
&lt;table>
</pre>

jayaddison:
The [`InBodyPhase`](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/html5lib/html5parser.py#L941) parser phase [creates and processes implied close-tag tokens](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/html5lib/html5parser.py#L1071-L1072) for a small number of [`stopNames`](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/html5lib/html5parser.py#L1065-L1067) elements.

This changeset updates the phase's logic so that processing of these implied tokens remains within the `InBody` parser phase.

This fixes #111 by avoiding a call to the phase-specific `InTablePhase.endTagOther` to handle the implied end-tag, which as a side-effect [disables the `insertFromTable` flag](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/html5lib/html5parser.py#L1820).  That flag must remain `True` in order for repositioning of fostered elements relative to a `<table>` tag to be performed correctly.


jayaddison:

Issue #445, opened 2020-02-26
Experiment with Cython


gsnedders:
If we can simplify some of the code slightly, we might be able to get some decent performance increase under CPython from using Cython (note that we almost certainly want to keep our codebase pure Python, so this means using Cython's pure-Python mode). It's probably at least worth experimenting here.

At least for `html5parser`, making decent progress is blocked on #272, and relatedly making `html5parser.HTMLParser.phases` not a dict (so they can all become struct references).

gsnedders:
#24 is the other part that would be significant here, I think.

PR #524, opened 2021-01-05
Compile html5lib with Cython

gsnedders:
Use Cython to make the parser quicker; see #445. This builds on top of #272.

This is a long way from ready to land, but shows potential. We probably also want to split this up so many of the earlier changes land first if they make performance sense without Cython.

The change to attribute representation especially might be of interest to #521 (cc @jayaddison).

There's also some API changes towards the end of the branch which we may well want to delay landing even beyond the rest of the Cython stuff.


html5lib/_inputstream.py
```diff
@@ -318,45 +360,98 @@ def charsUntil(self, characters, opposite=False):
...
 
-        while True:
-            # Find the longest matching prefix
-            m = chars.match(self.chunk, self.chunkOffset)
-            if m is None:
-                # If nothing matched, and it wasn't because we ran out of chunk,
-                # then stop
-                if self.chunkOffset != self.chunkSize:
+            while True:
+                # this really should be a slice of self.chunk, but https://github.com/cython/cython/issues/3536
```
gsnedders, line 400: ref https://github.com/cython/cython/issues/3536

html5lib/_inputstream.py
```diff
@@ -318,45 +360,98 @@ def charsUntil(self, characters, opposite=False):
...
+                for i in range(self.chunkOffset, self.chunkSize):
+                    with cython.boundscheck(False), cython.wraparound(False):
+                        c = self.chunk[i]
+                    if c > 0x7F and opposite or c <= 0x7F and (bitmap[index(c)] & bit(c)):
+                        cyrv += self.chunk[self.chunkOffset:i]
```
gsnedders, line 405: ref https://github.com/cython/cython/issues/3887 (IIRC, this was a significant perf issue)

jayaddison:
Thanks @gsnedders - I see that the attribute names/values are no longer accessed via tuple/list indexes (a nice improvement), and that the element attributes are initialized as attribute maps (rather than being converted into them at the end of the process).

Those are similar to the approach taken in #521 as well.  What's the general approach to review, quality/performance assurance and merge for each pull request?  It looks tricky to review this changeset as one combined change (given the size of the diff) - could we create a list of the proposed changes (with dependencies / ordering between them) and queue them up for review?

gsnedders:
> What's the general approach to review, quality/performance assurance and merge for each pull request? It looks tricky to review this changeset as one combined change (given the size of the diff) - could we create a list of the proposed changes (with dependencies / ordering between them) and queue them up for review?

Yeah, I was mostly just throwing this up as one PR so at least the branch exists somewhere that isn't my local disk!

Also not sure why it fails on 2.7 or 3.6 on CI; it doesn't fail locally!

PR #526, opened 2021-01-09
Use Python built-in str.lower in preference to asciiUpper2Lower character table translation

jayaddison:
This change uses Python's builtin `str.lower` to perform string lowercasing (typically used during case-insensitive element/attribute name comparison), and provides a nice small parsing performance benefit on Python (cpython 3.9.1).

Benchmark command used: `python benchmarks/bench_html.py parse --rigorous`

**Before**
```
html_parse_etree: Mean +- std dev: 208 ms +- 14 ms
```

**After**
```
html_parse_etree: Mean +- std dev: 200 ms +- 12 ms
```

This follows on from some [conversation in #520](https://github.com/html5lib/html5lib-python/pull/520#issuecomment-754098934) regarding comparison operations and lowercasing.


jayaddison:

Issue #525, opened 2021-01-08
Move away from Travis CI


gsnedders:
Probably to GitHub Actions?

While we're rewriting the config, we may well want to migrate to [cibuildwheel](https://pypi.org/project/cibuildwheel/) at the same time, especially before #445 happens (when we'll then need the infra to build non-universal wheels for releases cross-platform).

jayaddison:
GitHub actions is probably the most straightforward alternative at the moment, yep.

Could we build [`universal` wheels](https://wheel.readthedocs.io/en/stable/user_guide.html?highlight=universal#building-wheels)?  I think that'd allow building and publishing wheels without introducing additional build-time dependencies.

Thoughts on build configuration:

During migration, running `tox` is likely the simplest way to run tests under all of the configured [environment combinations](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/tox.ini#L2).  That would run all of the tests sequentially in each of the environment configurations and output a single pass/fail indicator in GitHub for the developer.

Perhaps it'd make sense during the migration to have the chosen continuous integration system run each of the configured environments as a separate build (i.e. individual process with pass/fail indicator)?

There are a reasonably large number of environment combinations already (21 by my count?), so we'd want to be careful not to create redundant duplicate work - especially if opening pull requests causes all those workloads to be initiated (resulting in expenditure of compute resources (and therefore energy :)) that it'd be better to conserve).

jayaddison:
Hrm, mapping GHA `strategy.matrix`-based jobs to single `tox` environments looks a little tricky due to the [combinatorial `envlist`](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/tox.ini#L2).

hugovk:
GitHub Actions was added in https://github.com/html5lib/html5lib-python/pull/527, and tests are also run on AppVeyor for Windows (this could also be moved to GHA, but that's another issue).

Also Travis CI hasn't been running for 10 months:

<img width="1040" alt="image" src="https://user-images.githubusercontent.com/1324225/161446169-18157d30-d93e-474e-92da-8c46963a3883.png">

https://travis-ci.org/github/html5lib/html5lib-python/builds

Please see https://github.com/html5lib/html5lib-python/pull/545 to remove Travis CI config and replace with GHA in docs.

PR #527, opened 2021-01-10
Add GitHub Actions workflow for tox

jayaddison:
Provides a migration path for #525.

- ~~The [`python debug.py` post-build step from `.travis.yml`](https://github.com/html5lib/html5lib-python/blob/2c19b9899ab3a3e8bd0ca35e5d78544334204169/.travis.yml#L24-L25) has not been migrated; GitHub Actions workflows support a [`continue-on-error`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) configuration flag at the job and step level, but I don't yet see whether this can be used to provide an equivalent 'post-job' command.~~
- ~~As of the initial version of this pull request, builds take place within a single GitHub Actions workflow that runs `tox` in parallel mode.  Build output for each Python environment is present in the workflow console, but it's not as human-readable as the grouping provided by Travis CI.~~


jayaddison:
See https://github.com/jayaddison/html5lib-python/runs/1677406878 for example GitHub Actions build output corresponding to c2171332634532d4ffd29e5edfe09bf6076fb279.

hugovk:
GitHub Actions provides 20 parallel workers, have you considered using a matrix to test each Python version in its own worker rather than all in parallel in a single worker? Should be much quicker.

hugovk:
But otherwise, big 👍  for moving from Travis CI to GHA.

jayaddison:
@hugovk Thanks!  Using a matrix strategy would seem ideal - the challenge is the [combinatorial `tox` `envlist`](https://github.com/html5lib/html5lib-python/blob/569e580d93f550033fde175cac1118ed26941332/tox.ini#L2), since the Python interpreter versions provided in GitHub Actions don't correspond to the `tox.ini`-defined interpreter names.

The [`tox-gh-actions`](https://github.com/ymyzk/tox-gh-actions) plugin is fairly widely adopted to solve more-or-less this problem, and it does that by reading a Python-version-to-environment mapping from the `tox.ini` file ([example](https://github.com/ymyzk/tox-gh-actions/blame/master/README.md#L39-L45)).

It's worth pausing and thinking about this change for a little while before merge, I think.  There may be a way to achieve the worker-per-environment goal in a different way - or perhaps it should be reported / discussed upstream.

jayaddison:
https://github.com/tox-dev/tox/issues/659 may be a relevant issue to track relating to this

jayaddison:
The latest commits re-attempt a matrix build strategy via GitHub Actions; the [generic `py` prefix for `tox` generative names should map to the 'current default Python interpreter'](https://tox.readthedocs.io/en/latest/config.html#tox-environments), so the generative factor is removed from `tox.ini` and replaced by the GHA YAML configuration.

This does have an unfortunate side-effect that it'll be trickier for developers with a local checkout of the repository to test the package against all Python versions by simply running `tox`.  Instead it will test against their default Python interpreter.  That doesn't seem perfect.

.github/workflows/python-tox.yml
```diff
@@ -0,0 +1,19 @@
+on: [pull_request, push]
+jobs:
+  build:
+    # Prevent duplicate builds for 'internal' pull requests on existing commits
+    # Credit: https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012
+    if: github.event.push || github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [2.7, 3.5, 3.6, 3.7, 3.8, pypy-2.7, pypy3]
```
jayaddison, line 10: NB for review: do `pypy-2.7` and `pypy3` respectively correspond to `pypy` and `pypy3` as previously listed in `tox.ini`?

jayaddison:
> GitHub Actions provides 20 parallel workers, have you considered using a matrix to test each Python version in its own worker rather than all in parallel in a single worker? Should be much quicker.

Ok, my mistake previously @hugovk; I overcomplicated things.  This _does_ look feasible after all, with one caveat/tradeoff regarding developer experience for local checkouts.

See https://github.com/jayaddison/html5lib-python/runs/1691598506 for an example multi-pipeline build via GHA using a matrix strategy

jayaddison:
... and perhaps that local checkout dev experience issue is now resolved too.  Each GHA worker now runs `tox -e py`, invoking the test suite for a single Python interpreter, as configured via the matrix worker environment list.

The remaining tradeoff/downside is that the environment list now exists in `tox.ini` (for local workflows) and is duplicated in the GHA YAML file.

https://github.com/jayaddison/html5lib-python/actions/runs/481474149 contains the latest build output as per commit 19cf7f750a74c95dad92c64b811fa859dcc4313b

hugovk (approved):

jayaddison:
Ping @gsnedders @jgraham @willkg

jgraham (approved):

Issue #537, opened 2021-11-04
Python 3.10 support


sanderbrauwers:
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3/dist-packages/pip/__main__.py", line 19, in <module>
    sys.exit(_main())
  File "/usr/lib/python3/dist-packages/pip/_internal/cli/main.py", line 73, in main
    command = create_command(cmd_name, isolated=("--isolated" in cmd_args))
  File "/usr/lib/python3/dist-packages/pip/_internal/commands/__init__.py", line 96, in create_command
    module = importlib.import_module(module_path)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/pip/_internal/commands/install.py", line 24, in <module>
    from pip._internal.cli.req_command import RequirementCommand
  File "/usr/lib/python3/dist-packages/pip/_internal/cli/req_command.py", line 15, in <module>
    from pip._internal.index.package_finder import PackageFinder
  File "/usr/lib/python3/dist-packages/pip/_internal/index/package_finder.py", line 21, in <module>
    from pip._internal.index.collector import parse_links
  File "/usr/lib/python3/dist-packages/pip/_internal/index/collector.py", line 12, in <module>
    from pip._vendor import html5lib, requests
ImportError: cannot import name 'html5lib' from 'pip._vendor' (/usr/lib/python3/dist-packages/pip/_vendor/__init__.py)
```

tp1050:
Same here
https://stackoverflow.com/questions/70132656/pip-fails-with-python3-10-python3-11/70132674#70132674

hugovk:
[...]

PR #542, opened 2022-04-03
Add support for Python 3.9 and 3.10 and 3.11

hugovk:
Closes https://github.com/html5lib/html5lib-python/issues/537.

Python 3.9 was released on 2020-10-05.

Python 3.10 was released on 2021-10-04.

Python 3.11 was released on 2022-10-24.

Let's test them and declare support.

### DIFF
--- a/.github/workflows/python-tox.yml
+++ b/.github/workflows/python-tox.yml
@@ -8,12 +8,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "pypy-2.7", "pypy-3.8"]
+        python: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "pypy-2.7", "pypy-3.8"]
     steps:
       - uses: actions/checkout@v3
         with:
           submodules: true
-      - uses: actions/setup-python@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/python-tox.yml
```diff
@@ -6,15 +6,18 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: true
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python }}
+          cache: pip
+          cachr-dependency-path: "requirements*"
```
jayaddison, line 20: A couple of details on this line:

- Typo in the key name: should be `cache-dependency-path`
- Perhaps it'd be worth bounding the wildcard expansion to `requirements*.txt`?

hugovk: Good spot, both done.

hugovk:
The Python 3.7 CI failure is unrelated and due to dependency changes. Please see https://github.com/html5lib/html5lib-python/pull/554 for a fix.

hugovk:
Updated to add support for Python 3.11.

The Python 3.5 and 3.6 failures are unrelated and because GitHub Actions no longer supports then because they're EOL (since 2020-09-13 and 2021-12-23).

PR #559, opened 2023-02-20
Restore ability to provide a tree builder name as string

### LINKED ISSUE #513: Docs out of date
In `HTMLParser.__init__`, `tree` is a class:

https://github.com/html5lib/html5lib-python/blob/f4646e6ed4eeb9780f67d2083d0c09c8fffbec53/html5lib/html5parser.py#L93-L97

However the examples in the API docs, which also ends up in the website, uses a string

https://github.com/html5lib/html5lib-python/blob/f4646e6ed4eeb9780f67d2083d0c09c8fffbec53/html5lib/html5parser.py#L105-L109

### LINKED ISSUE #558: HTML5Parser has incorrect behavior
On Version 1.1

The doc strings give this example, which would be great, if it worked, but actually it fails when `tree = "lxml"`

```py
def __init__(self, tree=None, strict=False, namespaceHTMLElements=True, debug=False):
        """
        :arg tree: a treebuilder class controlling the type of tree that will be
            returned. Built in treebuilders can be accessed through
            html5lib.treebuilders.getTreeBuilder(treeType)

        :arg strict: raise an exception when a parse error is encountered

        :arg namespaceHTMLElements: whether or not to namespace HTML elements

        :arg debug: whether or not to enable debug mode which logs things

        Example:

        >>> from html5lib.html5parser import HTMLParser
        >>> parser = HTMLParser()                     # generates parser with etree builder
        >>> parser = HTMLParser('lxml', strict=True)  # generates parser with lxml builder which is strict

        """
```

theRealProHacker:
We can simply propagate the `ValueError` raised in `treebuilders.getTreeBuilder`

fixes #558 
fixes #513


jvanasco:
Thanks!

PR #567, opened 2023-03-03
Constant phases

ambv:
This is #272 but rebased on top of current master.

I made cursory benchmarking of this and I see no difference between master and this branch.

| Python  | Tests executed | Time with PR | Time on master |
| ------- | -------------- | ------------ | -------------- |
| 2.7.18  | 28238          | 47.3s        | 47.2s          |
| 3.9.16  | 19062          | 13.4s        | 13.6s          |
| 3.11.2  | 23563          | 12.4s        | 12.4s          |
| pypy3.8 | 23563          | 44.3s        | 44.4s          |

Each timing is geomean from 30 runs. This is executed on Homebrew builds of all Pythons above on an M1 Max Macbook Pro. The slow result from pypy3 (v7.3.9) is because it's running in Rosetta (x86-64 emulation). The important thing is to compare the time on the PR with the time on master.


ambv:
When looking at the PR without whitespace changes, it's not that complicated. All tests pass, no performance regression.

I think this should go in as is. This will help us optimize, whether with Cython or with mypyc, as the result is a plain old `Phase` class, which I like 👍🏻

html5lib/html5parser.py
```diff
@@ -204,6 +191,9 @@ def mainLoop(self):
         DoctypeToken = tokenTypes["Doctype"]
         ParseErrorToken = tokenTypes["ParseError"]
 
+        type_names = {value: key for key, value in tokenTypes.items()}
```
jgraham, line 194: I suppose one could skip computing this unless `debug` is set, but probably not worthwhile.

jgraham (approved):

PR #574, opened 2023-10-17
tests: drop dependency on external mock module for newer python

### LINKED ISSUE #541: use unittest.mock instead of mock
Could you please consider to use unittest.mock at least as fallback, when mock is not installed?

eli-schwartz:
Starting python 3.3, this is in the stdlib.

Fixes #541

### DIFF
--- a/html5lib/tests/test_meta.py
+++ b/html5lib/tests/test_meta.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, unicode_literals
 import six
 try:
     from unittest.mock import Mock
-except:
+except ImportError:
     from mock import Mock
 
 from . import support
--- a/html5lib/tests/test_meta.py
+++ b/html5lib/tests/test_meta.py
@@ -1,7 +1,10 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import six
-from mock import Mock
+try:
+    from unittest.mock import Mock
+except:
+    from mock import Mock
 
 from . import support
 
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -6,5 +6,4 @@ pytest>=4.6.10,<5 ; python_version < '3'
 pytest>=5.4.2,<7 ; python_version >= '3'
 coverage>=5.1,<6
 pytest-expect>=1.1.0,<2
-mock>=3.0.5,<4 ; python_version < '3.6'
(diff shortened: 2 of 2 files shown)


eli-schwartz:
All things considered I'd rather wait until I have core maintainer review and then rebase once instead of potentially many times.

cclauss:
If you were a core maintainer, would you feel more comfortable about reviewing a PR with failing tests (current state) or passing tests?   https://ci.appveyor.com/project/gsnedders/html5lib-python/history

html5lib/tests/test_meta.py
```diff
@@ -1,7 +1,10 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import six
-from mock import Mock
+try:
+    from unittest.mock import Mock
+except:
```
cclauss, line 6:
```suggestion
except ImportError:
```
See PEP8 or
* https://docs.astral.sh/ruff/rules/blind-except
* https://realpython.com/the-most-diabolical-python-antipattern

eli-schwartz:
If I were a maintainer, and:
- my tests don't pass because the test runner claims there are fatal errors in the Windows registry
- the issue has been ongoing for a very long time


I would not appreciate external drive-by contributors pressuring other external contributors to feel guilty about my project which doesn't have passing tests in the first place.

But also, as it happens, yes, I absolutely would review PRs with failing tests! Even if the failing test is a flaw in the contribution itself.

I do this ***all the time***. Then I offer suggestions to the contributors, who may not be fully aware of deeply hidden quirks of the codebase that require specialized knowledge to handle correctly, knowledge which I can provide in order to aid those contributors in getting their PR into a mergeable state.

Sometimes I, *gasp*, merge PRs with failing CI because the PR fixes one CI error but not all the errors, and CI will never be fixed if you're not willing to fix issues one by one and instead expect it all to be fixed in one fell swoop and the task becomes so intimidating that everyone just gives up and it never gets fixed.

Sometimes I review a PR and say "lgtm as soon as CI passes, please rebase", so that contributors can feel confident that they're not being ignored, and that if they rebase the PR *now* then they won't be asked to rebase it once a month for the next year only to have it never merged.

In short, yes, I review PRs without regard to whether they have passing CI, because I do not see what passing CI has to do with anything at the review stage.

cclauss (approved):

ambv (approved):

PR #580, opened 2024-09-09
Only support python versions that receive security updates

Mic92:
Even debian oldstable has python 3.9. For internet-facing libraries it
is not secure for contributor to install unsupported python versions using C-based xml parser in order to test them.
Reducing the number of python versions will make maintenance and testing easier.


Mic92:
This will than also allow us in a follow up pull request to get rid of six as well as adding typing hints to the API.

setup.py
```diff
@@ -110,7 +105,7 @@ def default_environment():
           'six>=1.9',
```
gustavi, line 105: Could you also remove six has a dependency since Python 2 support is dropped?

PR #589, opened 2025-09-13
`setup.py`: fix version parsing on Python 3.14 (ast.Str removed)

### LINKED ISSUE #588: html5lib fails to build on Python 3.14 (uses removed ast.Str node)
setup.py uses ast.Str, which was removed in Python 3.14. As a result, attempting to build or install html5lib under Python 3.14 fails with an AttributeError, similar to below:
```zsh
       >   File "setup.py", line 93, in <module>
       >     isinstance(a.value, ast.Str)):
       >                         ^^^^^^^
       > AttributeError: module 'ast' has no attribute 'Str'
```

sookach:
Python 3.14 removes the ast.Str node type. String literals now appear as ast.Constant(value=str).
Update the AST check to accept both ast.Str (for older Pythons) and ast.Constant with a string value (for Python 3.8+), allowing html5lib to build successfully on Python 3.14 while remaining compatible with older version. Tested at desk with `pip install -e . `

Fixes: #588


atsampson:
This isn't right for pre-3.14 versions - the `version =` line fails because there's no `a.value.value` on older versions. Something like this works (tested on 2.7 and 3.11-3.14):

```
version = None
with open(join(here, "html5lib", "__init__.py"), "rb") as init_file:
    t = ast.parse(init_file.read(), filename="__init__.py", mode="exec")
    assert isinstance(t, ast.Module)
    assignments = filter(lambda x: isinstance(x, ast.Assign), t.body)
    for a in assignments:
        if (len(a.targets) == 1 and
                isinstance(a.targets[0], ast.Name) and
                a.targets[0].id == "__version__"):
            if hasattr(ast, "Str") and isinstance(a.value, ast.Str):
                version = a.value.s
            elif (hasattr(ast, "Constant")
                  and isinstance(a.value, ast.Constant)
                  and isinstance(a.value.value, str)):
                version = a.value.value
assert version is not None
```

sookach:
Thanks! Updated the PR.

reneleonhardt:
Could you fix appveyor? For example the flake8 formatting.
https://ci.appveyor.com/project/gsnedders/html5lib-python/builds/52890634

setup.py declares python_requires >= 2.7, but Python 3.13 is not tested in appveyor.

Python 3.7 is EOL since 2023-06-27, when will it be allowed to not stay compatible anymore?

> py --list
>  -V:3.8           Python 3.8 (64-bit)
> No suitable Python runtime found

Python 2.7 is EOL since 2020-01-01, when will it be allowed to not stay compatible anymore?
Microsoft Visual C++ 9.0 is 15 years old and not available anymore officially of course.

> running build_ext
> building 'lxml.etree' extension
> error: Microsoft Visual C++ 9.0 is required. Get it from http://aka.ms/vcpython27

https://web.archive.org/web/20190720195601/http://www.microsoft.com/en-us/download/confirmation.aspx?id=44266

PR #592, opened 2026-02-09
setup.py: handle missing pkg_resources import

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/266589


loqs:
@chenrui333 will you also be doing a pull for pkg_resources use in html5lib/tests/conftest.py?

setup.py
```diff
@@ -6,9 +6,12 @@
...
 from os.path import join, dirname
 from setuptools import setup, find_packages, __version__ as setuptools_version
-from pkg_resources import parse_version
-
-import pkg_resources
+try:
+    from pkg_resources import parse_version
+    import pkg_resources
+except ModuleNotFoundError:
+    from setuptools._vendor.packaging.version import parse as parse_version
```
loqs, line 13: @chenrui333 @grawlinson this does not work on Arch linux. The vendor namespace is an internal implementation detail that should not be relied upon. setuptools 82 that removed pkg_resources requires the packaging package so if pkg_resources is not provided by setuptools packing will be available not withstanding a broken installation.

grawlinson: Like @loqs has said; It is 100% better to just do `from packaging import parse as parse_version` rather than depending on an internal implementation which will change at some point in the future.

loqs: I think the issue can be avoided as the check is only for setuptools less than 18.5 and pkg_resources was removed in setuptools 82 so the import failure means there is no need to do the version check and it is enough to set pkg_resources to None:
```patch
--- a/setup.py
+++ b/setup.py
@@ -6,9 +6,11 @@ import sys
 
 from os.path import join, dirname
 from setuptools import setup, find_packages, __version__ as setuptools_version
-from pkg_resources import parse_version
 
-import pkg_resources
+try:
+    import pkg_resources
+except ImportError:
+    pkg_resources = None
 
 try:
     import _markerlib.markers
```

PR #599, opened 2026-06-25
Fix AssertionError when parsing malformed HTML (closes #568)

### LINKED ISSUE #568: Fuzzing reveals a number of parse errors
I'm the lead developer of Beautiful Soup, which has html5lib as an optional dependency. Over the past couple of years I've gotten a number of notifications from Google's oss-fuzz project about unhandled exceptions that actually turned out to be problems in html5lib. There wasn't much I could do with these errors, but now that it looks like html5lib maintenance is picking up, I can pass them on to you. (Sorry. :crying_cat_face:)

I've incorporated the fuzz reports into [the Beautiful Soup test suite](https://git.launchpad.net/beautifulsoup/tree/bs4/tests/test_fuzz.py), and [the test cases themselves are here](https://git.launchpad.net/beautifulsoup/tree/bs4/tests/fuzz), but here's a general picture of what problems I see. In each case, I believe just parsing the bad markup is enough to trigger the error.

**clusterfuzz-testcase-minimized-bs4_fuzzer-4999465949331456**

Markup: `b')<a><math><TR><a><mI><a><p><a>'`

Error:

```
self = <html>, node = <p>, refNode = None

    def insertBefore(self, node, refNode):
>       index = self.element.index(refNode.element)
E       AttributeError: 'NoneType' object has no attribute 'element'
```

**clusterfuzz-testcase-minimized-bs4_fuzzer-5843991618256896**

Markup: `b'-<math><sElect><mi><sElect><sElect>'`

Error:

```
    def resetInsertionMode(self):
    ...
            # Check for conditions that should only happen in the innerHTML
            # case
            if nodeName in ("select", "colgroup", "head", "html"):
>               assert self.innerHTML
E               AssertionError
```

**clusterfuzz-testcase-minimized-bs4_fuzzer-6241471367348224**

Markup: `b'ñ<table><svg><html>'`

Error:

```
self = <html5lib.html5parser.getPhases.<locals>.InTablePhase object at 0x7f8f405ad440>

    def processEOF(self):
        if self.tree.openElements[-1].name != "html":
            self.parser.parseError("eof-in-table")
        else:
>           assert self.parser.innerHTML
E           AssertionError
```

**clusterfuzz-testcase-minimized-bs4_fuzzer-6600557255327744**

Markup: `b'\t<TABLE><<!>;<!><<!>.<lec><th>i><a><mat\x00\x01<mi\x00a><math>><th><mI>chardeta\xff\xff\xff\xff<><th><mI><||||||||A<select><>qu?\xbemath><th><mie>qu'`

Error:

```
self = <html5lib.html5parser.getPhases.<locals>.InTableBodyPhase object at 0x7f8f4184ce00>

    def clearStackToTableBodyContext(self):
        while self.tree.openElements[-1].name not in ("tbody", "tfoot",
                                                      "thead", "html"):
            # self.parser.parseError("unexpected-implied-end-tag-in-table",
            #  {"name": self.tree.openElements[-1].name})
            self.tree.openElements.pop()
        if self.tree.openElements[-1].name == "html":
>           assert self.parser.innerHTML
E           AssertionError
```

Also reported to me recently was the issue that was reported to you as issue #557.

gaoflow:
## Summary

Fixes two `AssertionError` crashes triggered by malformed HTML during a regular (non-fragment) parse, originally reported in #568 via Google's oss-fuzz / Beautiful Soup.

### Bug 1 — `resetInsertionMode` crashes on `<select>` inside foreign content

When a `<select>` element appears in the open-elements stack inside a foreign-content element (e.g. inside `<math>`), the subsequent call to `resetInsertionMode` encountered the element name in the `("select", "colgroup", ...)` guard on line 369–370 and raised an `AssertionError`:

```python
b'-<math><sElect><mi><sElect><sElect>'
# → AssertionError (html5parser.py resetInsertionMode line 370)
```

Per the WHATWG HTML5 parsing spec (§8.2.8.1 "reset the insertion mode appropriately"), `<select>` is **not** restricted to the innerHTML case — the correct mode is `"inSelect"` (or `"inSelectInTable"` if a table ancestor is present). The fix removes `"select"` from the asserted set so it falls through to the existing `newModes` lookup.

For `"colgroup"`, `"head"`, and `"html"` (which genuinely should only appear during fragment parsing), the hard `assert` is replaced with a `continue` so the loop seeks a better ancestor instead of crashing.

### Bug 2 — `InTablePhase.processEOF` crashes when `<html>` is current node

Malformed markup such as `ñ<table><svg><html>` can leave `<html>` as the current node while the parser is in "in table" mode outside of an innerHTML context. The old code asserted `self.parser.innerHTML` at that point:

```python
b'\xc3\xb1<table><svg><html>'
# → AssertionError (html5parser.py InTablePhase.processEOF line 1699)
```

The fix emits a `parseError` for the non-innerHTML path and then lets the parser stop cleanly, consistent with what the spec requires on EOF.

## Before / After

```python
import html5lib

# Both raise AssertionError before this PR:
html5lib.parse(b'-<math><sElect><mi><sElect><sElect>')
html5lib.parse(b'\xc3\xb1<table><svg><html>')

# Both return a document tree after this PR (with appropriate parse errors recorded).
```

## Test plan

- [x] `html5lib.parse(b'-<math><sElect><mi><sElect><sElect>')` — was `AssertionError`, now returns document
- [x] `html5lib.parse(b'\xc3\xb1<table><svg><html>')` — was `AssertionError`, now returns document
- [x] Normal HTML, `<select>` forms, `<table>`, and MathML documents continue to parse correctly

---

This pull request was prepared with the assistance of AI, under my direction and review.


binggao1230:
Pushed `d9ac326` adding focused regression coverage for both malformed inputs from the PR description.

Verification run locally:
- `PYTHONPATH=$PWD uv run --with-requirements requirements-test.txt --with 'setuptools<80' python -m pytest html5lib/tests/test_parser2.py -q`
- `PYTHONPATH=$PWD uv run --with-requirements requirements-test.txt python -m flake8 html5lib/html5parser.py html5lib/tests/test_parser2.py`
- `git diff --check`

I also checked the current AppVeyor failure. It looks unrelated to this parser change: the Python 3.7 jobs fail before tests at `py -VV` with “No suitable Python runtime found”, and the Python 2.7 optional job fails while installing `lxml 5.0.2` because Windows cannot build it without VC++ 9.0.

